PR #4, opened 2015-03-06
Removed units warning dialog

tjlang:
If the radar file for some reason lacks unit labels for certain variables, ARTview gives a popup dialog warning before plotting. This is OK the first time, but to receive this tilt after tilt is really annoying. I have commented that out.

### DIFF
--- a/artview.py
+++ b/artview.py
@@ -587,6 +587,7 @@ class Browse(QtGui.QMainWindow):
         '''Captures a selection and redraws the new field'''
         self.field = nombre
         self._initialize_limits()
+        self.units = None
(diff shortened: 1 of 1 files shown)


tjlang:
Added buttons to set plot Title and plot Units, if you don't want the Py-ART defaults. If you delete the text in the button dialog, the Title and Units will revert back to Py-ART defaults. This pull request supersedes my previous one from today.

screen shot 2015-03-06 at 1 12 58 pm
. If you enter nothing

nguy:
@tjlang , I've merged you fork and pushed the changes to the master.

PR #7, opened 2015-03-24
Made directory argument optional

tjlang:
It annoyed me to have the directory be a required command-line argument, so I made it optional. I did this a while ago but since you have been updating the program lately I thought it would be good to get it into the master branch.

Now you can just do "python artview.py" and it will bring up the directory menu to choose the file.


tjlang:
Er, never mind looks like this was already done?

PR #11, opened 2015-03-31
Click on a point and get the value at the point 

pfhein:
I modified onPick to output the value at the point as well as the point's azimuth and range.  (It looks like there are 100s of changes but reality is only onPick was changed.  I copied you master over and did not go through github like I should have.)


nguy:
@pfhein Nice work!  I've implemented the changes and pushed a new version to github.
Minor changes in display, but that is all.
Let me know if this is satisfactory so we can close the issue.

pfhein:
Looks good, Nick.  I guess I would like one more digit on the range so 
that it does not truncate to 10's of meters.  I like seeing the "5" on 
my 125 meter gate spacing in my test files.  That is my only complaint.
Paul

nguy:
That's fair.  I updated that for the range display, but kept the others more for space.
It is in the master code and will be updated in the next go-round.

PR #13, opened 2015-05-02
Small improvements

gamaanderson:
several small improvements, including: 
1. buttons were blocking arrow key, so need to set policy No.Focus
2. add drop-down menus for tilts and fields
3. add field radio menu based in tiltButtonWindow, open with the drop-down menu in item "Field Window"
4. update field and tilts menus when radar changes.

The alteration are disperse in all code, hope it don't conflict with any thing you are doing. 
Notice we have now 3 controls over tilts: drop-down, radio and arrow. I test them and are all working as well as automatic updating in all circumstance I could think of.


nguy:
@gamaanderson - These changes are great.  I went ahead an merged these and figure we can modify the code as needed.

I like the fact that both displays can be changed.  But I think this should be a global type of change from the tilt button (meaning changing all open displays) vs. the drop down only changing the local display selected.

I also made some changes and pushed them over.  There are some big and small modifications:
Connecting functionality to new interface.
Display redesigned to simplify buttons.
Added a display options and added title, units, colormap and limits adjustment to this drop down menu.
Added Field Window and Tilt Window as selectable menu.
Added Tools drop down menu for zoom/pan and restore default.
Eventually can add custom tools to this area.

Cleaned up the code a little by modularizing where possible and adjusting comments.

Added a new limits external class, but this is not fully operational.


I like the Menus you created, but I find the Window selection to be awkward.  Do you think drop down menus like the two new additions would be better?

nguy:
I also tried to "clean up the code" by moving some aspects into modules and organizing a bit.  Hopefully this gives an idea of what I am thinking.  

One key component I hope is the toolbox.  This will be the place where a user can select tools and even load a local tool for testing. I'm open to suggestions on this.

There are a few problems.  The **Adjust Limits** under the Display Options is not sending a signal back for update.  The new values are being recorded, but for some reason the update is not occurring. I must be missing something that you did in the new field and tilt scripts...

Also, when a tilt is selected by the menu, it changes both displays.  I think it should only change one display.  A separate global change would be nice to add as well, where you change a tilt and all displays are changed.

nguy:
@tjlang and @pfhein check out the new uploads.  What do you guys think of the structure?

gamaanderson:
@nguy thank you for the alteration, in special I'm always in favor of cleaning up the code, formatting according to PEP8, removing unused inherited parts and historical comments. Adding docstrings so that we can create an manual afterward is also important, since this project has both a software part (the `execute` file) and a library part (the plot.py etc.). This goals are all being achieved gradually, so it is all good. 

As for the code alteration I have lots of comments, and not all are favorable. I will post the smallest points here and open issues to discuss the other. I will also make comments to the commit, as to call attention to small things in the programing.

gamaanderson:
1. I work with file where the standard field is "DBZH", so I couldn't execute the code without make alteration. Function `_initialize_limits` should have a standard result, so that the program don't collapse and I can then change the limits manually. 

2. Buttons self.dispButton and self.toolsButton are getting focus blocking the arrows keys, one must set `.setFocusPolicy(QtCore.Qt.NoFocus)` in them.

3. The problems you are having with **Adjust Limits** are not in `limits.py`, but in `plot.py` it is using `self.limits` and not `self.Vlims`, so you update is just not being used. And moreover, with starting `Ui_LimsDialog` you are giving back `self.limits` ovewriting `self.Vlims`. Those are redundant variables, but wait for the topics on global variables, before deciding to change this.

4. I don't think the Window selection is awkward, but I also don't have anything against the new button style. This is your call. If you open an issue as a "to do" point, I can eventually implement whatever decision you take.

gamaanderson:
It is already late, I will open some issues tomorrow.

nguy:
@gamaanderson Thanks for the comments.  I'll respond some here and in each open issue.

The standard field right now is admittedly a "hack".  I had not come up with a clever way to do this, but I've added the 'DBZH' field into the ```python _check_default_field``` function - maybe there is a better way of doing this...  There are so many possibilities out there and I only listed the ones that I've dealt directly with or have been asked to add.  If you know of more we can add them.

I added the ```python .setFocusPolicy(QtCore.Qt.NoFocus)``` to self.dispButton, it was already present for the self.toolsButton

Let's keep the window selection for now.  I do like the new button style somewhat better, but this may be personal preference.  Maybe we can get some feedback from others on this before making any changes.

So one general comment about the code is that I really like the two window option as a "default".  So I may disable most of the main menu for a while but keep in the code in case we want to add this back later.  We'll need to decide on how we want to deploy the program, i.e. just have one executable like **execute_two.py** or give the user more control.  And then how to do that.

gamaanderson:
Ok
I'm also interested in some more feedback

I'm in favor of giving the used more control, nevertheless we should supply a standard **execute.py**.

nguy:
I just pushed some some incremental updates.

nguy:
I agree with the standard execute.  My preference is something similar to what is in there now.  Maybe some modifications as needed.

pfhein:
I have not had time to take a close look but I think things are looking good.  You asked about the structure.  As I said in another comment on a different issue, I would separate out the toolsBox and put that and its classes in a subfolder so to keep things clean and simple, and to make it easier for someone to add their special tool without overly complicating things.

nguy:
Thanks @pfhein , I do agree and I think we'll do this down the road.  In the short term, I've lumped a few together that inherent to passing along to the plot.  Though this may be split out.

We'll have to get a ```python setup.py``` working before we can have a separate folder, but ultimately this will happen I feel.

nguy:
Okay @gamaanderson @pfhein @tjlang the latest updates are now committed.  I likely won't be able to do much with the code for a short while now, but I think it has progressed very well in the last few days!

@swnesbitt - If you get a chance try the new flavor instead of ```python artview.py -d somedirectory``` you can use ``` python execute_two.py -d somedirectory``` to get to the **new** version.

gamaanderson:
Ok, if this means, at least by now, the time of big code changes is completed, I will start handling isolated issues and creating PR for those.

nguy:
Have at it @gamaanderson .  Oh and could you send me an email nick.guy@uwyo.edu?

Issue #17, opened 2015-05-05
Class Initialising Variables


gamaanderson:
I open this to discuss if and when class should initialize V variables. I usually don't like this, as it is giving class to much power, and therefore multiple classes can conflict with each other over this. 

This is happening in two places now: `Menu` initialize `Vradar` and `Display` initialize `Vlims`. The first one, I did, because the main function of `Menu` is to govern over `Vradar`, so it is not that bad that it open it. Nevertheless I believe this should be done some what differently, not sure exactly how.

gamaanderson:
This is a small issue, so I would wait the new style to settle before handling it.

nguy:
Sounds good.  Let's see how things fall out.  But my initial (ignorant) feeling is that if this is controlled in the execute functions, it may not be detrimental?

gamaanderson:
This is not a ignorant feeling at all, it is a pretty real concern, that is why it need some thought before doing.

gamaanderson:
Before closings this issue, I would like to make a question. @nguy has suggested a graphical interface for starting a component, in especial a `Display`. I believe this could work fine as a static method inside every component. There is just one complication, the shared variables (V Variables). There are other solution for that, but one would be have all shared variables as optional arguments.
However I don't like, AT ALL, of `Display` initializing `Vradar`. Are you as opposite to that as I am? As I said the are other options to that, for instance a more complex starting interface or make Display work (that is: not crash!) with `Vradar.value == None` and count on issue #14 to connect to a valid radar.

nguy:
I have actually changed my mind about a GUI for starting a component (for now).  I don't see the benefit in this.  I believe we should just have a default startup mode (same or similar to what we have now) and let the user add instances (another Display window for example) via the main menu.  

I don't think that Display should initialize Vradar either.  Right now the Vradar instance is started in ```python menu.py``` and I do like this structure.  If there was a main abstract class that contained the whole app, this would allow the Vradar instance to be passed easily (I believe) to a new Display.  In addition, we would like to be able to ultimately edit the data (i.e. via selectable regions and automated functions) and this would likely make that coding a bit more straightforward.

One question I would have from the above is: Do we need to make Display not crash when ```python Vradar.value == None```? We could put a check in that passes when this is the case?

gamaanderson:
I'm not quite sure about the implementation, but it would be kind of interesting let the Display blank while waiting for a radar. I think it wouldn't be difficult to get that.

nguy:
[...]

Issue #19, opened 2015-05-06
Conflicting changeValue


gamaanderson:
I have realized that it can happen that several V variables are changed at the same time, e.q. Vradar and Vfield because the new radar doesn't have the previous field. I'm not fully sure what would happen then, but I believe _update_plot() would be called twice, this would be wasted time to say the least.

I will present my thoughts bellow.

gamaanderson:
At first I thought to hide the alteration of field calling

    self.Vfield.value = new_field

But that is really wrong, since other classes may be holding just the Vfield but not the Vradar, they how then not be informed of the change. I decided then to make two kinds of value change: a soft and a strong one and is responsibility of the Display to react to the right one. There are two ways of doing that:

1. Variable have two signal "ValueChanged" and "StrongValueChanged", normally both are emitted but in the case of a soft update just the first is emitted. Display on the other hand would just react to "StrongValueChange"

2. There is still just one signal, but V variable send and extra argument `strong=True/False`. Then display would do the following:

        def NewField(self, variable, value, strong):
            # do some thing 
            # ...
            if strong:
               self._update_plot()

Comments?

nguy:
I agree that calling ```python _update_plot()``` more than once is wasted time and/or resources.  I admit that I haven't checked the signals coming back into the Display instance.  I'm not sure I understand the difference between strong and not strong.

I just went back and looked at the code again.   In cases like say the ```python tilt.py```, is it necessary to have Vradar called?  If you comment that section out, there is seemingly no affect on the Display.  Can this be safely removed then?

I assumed that the .change takes the change and updates the Variable instance in the Display.  Is this the case?  And if so, why not drop the Vradar altogether?

Finally, are we talking about two different things here?

gamaanderson:
Yes, we are probably talking about different things. So I will not comment your questions, since I'm not sure if I understood.
I'm talking about, suppose for safety `menu.py` around line 313 is updated to the following

    radar = pyart.io.read(self.filename)
    if self.Vfield.value not in radar.fields.keys():
        self.Vfield.change('reflectivity')
    self.Vradar.change(radar)

this is just an example, menu don't have Vfield and this check should be done by Display, I believe. But any way, in this case, plot is update twice. I would like to do the following

    radar = pyart.io.read(self.filename)
    if self.Vfield.value not in radar.fields.keys():
        self.Vfield.change('reflectivity',strong=False)
    self.Vradar.change(radar,strong=True)

[...]

Issue #14, opened 2015-05-05
Global Variables


gamaanderson:
It was suggested to include of having the drop-down tilt menu changing local tilt and the TiltWindow changing the global tilt. I open this issue to discuss this.

As for me, this is not possible, not should it be possible. I will explain why is that bellow.

gamaanderson:
I'm not saying I don't like the idea, but it have severe consequences for the code.

The current structure is based in the idea that **everything is local** and should work locally, there is not global variable and global structure is just the sum of local ones. Local structures can share or not variables, that is all.

This reflect in the fact that the `Variable` class do both: hold the value and emit `"ChangeValue"`signal. To make the kind of selectable alteration suggested one must separate this two functions in `self.tilt`and `self.Vtilt`. This is the first problem, increase code complexity.

But my fundamental problem is that having global variables implies we are thinking in a specific global structure, this result that alternative configurations will must probably suffer with unwanted behavior. Also, programing turns more difficult, since with every variable I will have to care where to use the local one and where to use the global/shared one, this reflect in time. And the most time we spend in this is the less time spent in functionalities.

gamaanderson:
As an alternative for this, I would be in favor of a class that connect/disconnect variables of other classes. I, however, didn't have a good idea of how to perform this, but I believe classes show need to inform what are the sharable variables. Uniform naming of those variables would also be important.

nguy:
@gamaanderson I'll need to wrap my head around this for a bit.  I'm new to this particular style with ```python Variable``` class.  I think I see how this works now, but I don't have a full understanding of the effects quite yet.  Your comments have been most helpful.

I agree that this functionality is a challenge in our current setup.  I like the connect/disconnect idea.  Maybe for now we can form a better plan and comment here as ideas strike us and tinker around.  I think the current way is not detrimental.

Let's leave this open, but I'll concentrate on other parts first before we tackle this issue.  Sound good?

gamaanderson:
Sound perfect

gamaanderson:
Just a heads up, I'm working in the connect/disconnect idea in a [branch](https://github.com/gamaanderson/artview/tree/issue_%2314), and its is becoming a big change. Everything works the same way but I had to create a abstract Class from which all components (Menu, Display, FieldWindow etc.) are derived from.  Components must also inform which variables are the shared/sharable. Its turning good, when I feel it is mature I will open a PR. Till then continue pushing commits as you like, I will make sure of keeping the branch up-to-date.

[...]

PR #20, opened 2015-05-07
make Vradar optional for Menu and Vlims optional in Display

gamaanderson:
Issue #17
Turn some V variable into optional arguments, those are initialized internaly.
In **execute_two.py** I accessed Vradar using `MainMenu.Vradar` instead creating function  `Main.Menu._getVradar()`. I don't think we need a get function here, do you agree?

### DIFF
--- a/execute_two.py
+++ b/execute_two.py
@@ -24,18 +24,18 @@ class make_MainDisplay(object):
 
 DirIn,field=parser.parse(sys.argv)
 
-Vradar = Variable(None)
+
 
 
 app = QtGui.QApplication(sys.argv)
 
-MainMenu = Menu(Vradar, DirIn, name="Menu") #initiate Vradar
+MainMenu = Menu(DirIn, name="Menu") #initiate Vradar
+Vradar = MainMenu.Vradar
 
 Vtilt = Variable(0)
 Vtilt2 = Variable(0)
-Vlims = Variable(None)
-plot1 = Display(Vradar, Variable(field), Vtilt, Vlims, name="Display1", parent=MainMenu)
-plot2 = Display(Vradar, Variable(field), Vtilt2, Vlims, name="Display2", parent=MainMenu)
+plot1 = Display(Vradar, Variable(field), Vtilt, name="Display1", parent=MainMenu)
+plot2 = Display(Vradar, Variable(field), Vtilt2, name="Display2", parent=MainMenu)
 
 app.exec_()
 
--- a/menu.py
+++ b/menu.py
@@ -10,11 +10,12 @@ import os
 from PyQt4 import QtGui, QtCore
 
 import common
+from core import Variable
 
 class Menu(QtGui.QMainWindow):
     '''Class to display the MainMenu'''
 
-    def __init__(self, Vradar, pathDir, name="Menu", parent=None):
+    def __init__(self, pathDir, Vradar=None, name="Menu", parent=None):
         '''Initialize the class to create the interface'''
         super(Menu, self).__init__(parent)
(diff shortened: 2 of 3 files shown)


menu.py
```diff
@@ -29,7 +30,10 @@ def __init__(self, Vradar, pathDir, name="Menu", parent=None):
         self.LaunchApp()      
                 
         # Show an "Open" dialog box and return the path to the selected file
-        self.showFileDialog()
+        # Just do that if Vradar was not given
+        if self.Vradar is None:
+            self.Vradar = Variable(None)
+            self.showFileDialog()
```
nguy: I like this better than having the instantiation in in the execute script.

execute_two.py
```diff
@@ -24,18 +24,18 @@ def __init__(self, menu, plots=None):
...
 
 Vtilt = Variable(0)
 Vtilt2 = Variable(0)
-Vlims = Variable(None)
-plot1 = Display(Vradar, Variable(field), Vtilt, Vlims, name="Display1", parent=MainMenu)
-plot2 = Display(Vradar, Variable(field), Vtilt2, Vlims, name="Display2", parent=MainMenu)
+plot1 = Display(Vradar, Variable(field), Vtilt, name="Display1", parent=MainMenu)
+plot2 = Display(Vradar, Variable(field), Vtilt2, name="Display2", parent=MainMenu)
 
 app.exec_()
```
nguy: I think it's a good idea to clean this script up.  
A thought: Something we might consider is building and function that would return Variable instances such as Vtilt (and potential other in the future) and use these for Display calls.  We could have a default method and then it might be scalable for the future.  I'm thinking of adding to the main menu a **Create Display** menu that pops up an additional window.  Then this would remove the need for multiple execute scripts or we could then call this the execute script.  Thoughts?

gamaanderson: This would be an interesting thing to have, however I tend to believe as complex and powerful a graphical interface could be, it will never be as powerful and versatile as text. My idea is that powerful user will still work with execute scripts, so as long as we can satisfies less powerful users with a standard execute there is no need to spend time in this.

That being said, I will create a graphical component to control how V variable are shared (issue #14). This could be extended in the future to handle the components it self.

nguy: Hmmm, thinking as something later that we rename **execute_two.py** to **execute.py**.  This script could continue to be basic.  I think the create display would be good to have though in case the less powerful user closes a window on accident (I'm guilty of this sometimes) and wants to bring it back up.

gamaanderson: Ok, this is a good point, I will keep it in mind.

nguy:
Just ran the code and maybe I failed to notice this before but self.fieldnames is not defined anywhere so there is a error in **_check_default_field** in plot.py.  If you are going to make more changes soon would you stick this in the __init__?  Otherwise, I'll do at a later date.  The code still functions as is so I consider this minor for the time being.

gamaanderson:
I will add that to my next PR, some time later today

nguy:
Thanks!

gamaanderson:
I'm not getting this error, `self.fieldnames` is being defined in `self.NewRadar()` that should be executed priory to any call of `self._check_default_field()`.
There is another error "  'Display' object has no attribute 'ax'  " that was caused by this PR, however I already have a solution using issue #19. 
Anyway, the code is working, I will keep an eye open to see if this turn back up and how to make  Display Start-up less vulnerable to such things.

nguy:
Hmm weird.  I still get the above stated error, but I can't reproduce the 'ax' error.
Will have to keep an eye on this.

PR #22, opened 2015-05-08
Issue #19

gamaanderson:
Issue #19. Create two kinds of value change, a `strong=True` and `strong=False`. The first is the standard one, and the second is a signal to suppress overly costly updates. Its the component responsibility of correctly respond to the difference. Give it a look, in how it is being used in `plot.py` lines 67 and 395

### DIFF
--- a/core.py
+++ b/core.py
@@ -7,14 +7,37 @@ Class instance to create Variables and establish change signals.
 # Load the needed packages
 from PyQt4 import QtGui, QtCore
 
-class Variable (QtCore.QObject):
-    def __init__(self,value=None):
+class Variable(QtCore.QObject):
+    def __init__(self, value=None):
         """Class that holds a value, changing that with change() emits a signal"""
         QtCore.QObject.__init__(self)
-        self.value=value
+        self.value = value
 
-    def change (self,value):
-        self.value=value
-        self.emit(QtCore.SIGNAL("ValueChanged"),self,value)    
+    def change(self, value, strong=True):
+        """
+        Change the Variable value and emit 'ValueChanged' signal
+
+        Parameters
+        ----------
+        value : 
+            New Value to be assigned to the variable.
+        strong : bool, optional
+            Define if this is a strong, of a soft change. This is a some what
+            subjective decision: strong is default, a soft change should be
+            used to indicate to the slot that this change should not trigger
+            any costly computation. Reasons for this are: this variable is
+            likely to change again shortly, another more important variable
+            is being changed as well etc.
+            .. note:
+                Defining how to respond to strong/soft changes is
+                responsibility of the slot, most can just ignore the
+                difference, but the costly ones should be aware.
+        
+        Note
+        ----
+        The arguments of the emitted signal are (self, value, strong)
+        """
+        self.value = value
+        self.emit(QtCore.SIGNAL("ValueChanged"), self, value, strong)    
 
 
--- a/core.py
+++ b/core.py
@@ -25,9 +25,9 @@ class Variable(QtCore.QObject):
             Define if this is a strong, of a soft change. This is a some what
             subjective decision: strong is default, a soft change should be
             used to indicate to the slot that this change should not trigger
-            any costly computation. Reasons for this are: this variable is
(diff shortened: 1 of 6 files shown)


core.py
```diff
@@ -7,14 +7,37 @@
...
+                Defining how to respond to strong/soft changes is
+                responsibility of the slot, most can just ignore the
+                difference, but the costly ones should be aware.
+        
+        Note
+        ----
+        The arguments of the emitted signal are (self, value, strong)
+        """
+        self.value = value
+        self.emit(QtCore.SIGNAL("ValueChanged"), self, value, strong)
```
nguy: Would it be useful to include an explicit example (lines of code) to show the costly change type, in case future developers want to add a function?

core.py
```diff
@@ -7,14 +7,37 @@
...
+    def change(self, value, strong=True):
+        """
+        Change the Variable value and emit 'ValueChanged' signal
+
+        Parameters
+        ----------
+        value : 
+            New Value to be assigned to the variable.
+        strong : bool, optional
+            Define if this is a strong, of a soft change. This is a some what
```
nguy: Should be "or a soft change"

plot.py
```diff
@@ -774,12 +778,12 @@ def _pass_lims(self):
...
         self.Vlims.change(self.limits)
              
-    def NewLimits(self, variable, value):
+    def NewLimits(self, variable, value, strong):
         '''Retrieve new limits input'''
         #self._pass_lims()
         print "In NewLims"
     
-    def NewRadar(self, variable, value):
+    def NewRadar(self, variable, value, strong):
```
nguy: It looks like NewRadar and NewLimits are outdated after ZoomPan update.  Just making a note so that I hope to remember to delete in a future push.

nguy:
@gamaanderson - You mention the "practical use" in the commit documentation.  Can you give some more detail on that?  Thanks!

core.py
gamaanderson: Yes, that would be useful, but I would ask you to do it. Not that I don't want, but as you is the one that have the experience of trying to understand how it works, I believe you would give a more useful/clear explanation.

core.py
gamaanderson: yes, it should

gamaanderson:
the use is is `plot.py`line 67. I know Vlims is a deprecated variable, but I made it a optional variable. In the case it is not given I already need to set as V Variable in line 51, but will just know its value in line 67 and this is after the first plot, so to avoid re-plotting I make a soft change.
The best would be not need to do that, but as I need and its a deprecated variable it is a good solution.

core.py
nguy: Okay, no problem.  I'll put this on my list.  When I clean up the some other bits and pieces (e.g. the ZoomPan below) I'll do this as well.

field.py
```diff
@@ -64,11 +64,11 @@ def SetFieldRadioButtons(self):
         
         self.NewField(self.Vfield, self.Vfield.value)  # setChecked the current field
```
nguy: L65 - Should this read ```python self.NewField(self.Vfield, self.Vfield.value, False)``` ?
This seems like a case where a weak change would be applicable?

gamaanderson: yes, this passed unnoticed. I did notice it in `tilt.py` and there I set as True, since it is intended to do the same as an ordinary change, i.e. check the radio button.
However, as I said, the difference between strong and weak is somewhat subjective. To avoid that, I chose "avoid costly computation" as the ultimate purpose of a weak change. As this is a direct call and `field.py` don't make any difference between strong and weak, the best would be to stay with the standard  `True`.

nguy: Okay, I didn't think of it like that.

gamaanderson:
One more thing, that error I was getting on  'Display' object has no attribute 'ax' was solved by that line 67 (the practical use) I was talking about.

nguy:
Very nice.  The issue I was having seems to be fixed as well.  Check out the latest push coming shortly.  I corrected some of the verbage and added a lot of documentation.

Note that the NewLims and NewRadar are actually not defunct and therefore retained.

Issue #3, opened 2015-02-20
ppi/rhi check too restrictive


tjlang:
ARTview appears to only consider the possibility that valid radar.scan_type attributes can be 'rhi' or 'ppi'. In fact, valid, plottable scan types include 'sector' and possibly other names. I changed the 'ppi' test in the #CreatePlot section from
if self.radar.scan_type == 'ppi':
to 
if self.radar.scan_type != 'rhi':
And I was able to plot up my 'sector' scan_type files just fine. I recommend more flexible checking for PPI vs. RHI style scans, or to just remove the tests and let ARTview fail with a warning if the scan type is really odd. 

tl;dr - Right now, to avoid the possibility that the program might fail on an unusual volume, the program is failing on perfectly valid volumes.

nguy:
@tjlang Thanks for the input! 
Paul H also had trouble with a surveillance scan, with what looks like the same reason.

I took your suggestion and it seems to work well with sector files I have locally.  This should also take care of surveillance scans.  I also changed another section to reflect this looser structure.

nguy:
@tjlang I think that fix worked for you.  Paul was having the same issue and the fix worked for him.

tjlang:
Looks to be fixed, agreed.

nguy:
Thanks for checking this @tjlang

PR #26, opened 2015-05-13
Issue #14

gamaanderson:
This is a major PR, push all changes you have so that I can merge them before the PR.
Tree major alterations:

1. Create Abstract component class from which all other derivate
2. Create ComponentsControl, I believe the use is intuitive, but suggestion are always welcome
3. Add layout and components menu in MainMenu, take a look in how it works, you may need to increase MainMenu window to see it. Also play with lines 36 and 37 in `execute_two.py`, comment each one of them out and see what happen.

### DIFF
--- a/component_control.py
+++ b/component_control.py
@@ -15,6 +15,8 @@ class ComponentsControl(core.Component):
     The user select 2 components from a list and a radio menu is added for
     every common sharable variable. The variable in the second one can be
     connect or disconnected from the first one.
+
+    This is a powerfull Component, multiple instances may conflict
     '''
 
     def __init__(self, components=None, name="ComponentsControl", parent=None):
--- a/component_control.py
+++ b/component_control.py
@@ -9,6 +9,7 @@ from PyQt4 import QtGui, QtCore
 from functools import partial
 
 import core
+import common
 
 class ComponentsControl(core.Component):
     '''Class instance for control variables shared between components.
@@ -18,6 +19,11 @@ class ComponentsControl(core.Component):
 
     This is a powerfull Component, multiple instances may conflict
     '''
+
+    @classmethod
+    def guiStart(self):
+        val, entry = common.string_dialog("", "Component Name", "Name:")
+        return self(name=val)
 
     def __init__(self, components=None, name="ComponentsControl", parent=None):
         '''Initialize the class to create the interface'''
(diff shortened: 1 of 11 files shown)


nguy:
@gamaanderson The master should be up-to-date.

gamaanderson:
then it can be merge

gamaanderson:
Just add commit showing how to add plugins (user defined components). Its just preliminary, it need some adjustments.

nguy:
I'll get to this soon.  Busy end of the week for me.  Looking forward to browsing through the changes. Thanks!

gamaanderson:
ok, take your time.

nguy:
Okay, I had a chance to float through this a bit this morning. This is an incredibly powerful tool.  I'll comment below on various things before merging.
Let me make sure I understand a couple of things.  
The 'Layout' menu grows with each component that you add? I see that removing the default (Components) disables that menu.  Is there a way to regenerate the menu?

1. I don't see much difference when commenting in/out L36/37. As far as I could tell the operation was the same? Okay after a little  more playing around L36 omits the layout menu from building.  It also results in the Components(?) window being on top and not buried (item #3).

2. I think we will need to rework the wording to explain what is happening here.  It took me a minute to figure out and I think someone with no experience of diving into our code and conversations will be lost.

3. Currently this window gets buried by the Display plots.  We should find a way to put it to the front or different location to make sure people know it is there.  Maybe this should even be callable via the Mainmenu (maybe that is what you did there)?

4. I like how this will allow the user to add their own component. I think we will need to be vigilant in documentation for this to be used properly (even with advanced users) just because it is a unique interface.

5.

gamaanderson:
yes, the layout menu grow with every new add! once empty it also reopen with a new add, for instance using the components menu

1. L36 put ComponentsControl inside MainMenu, instead of being a separated window. L37 add the possibility of re-opening ComponentsControl using the components menu in the menubar of MainMenu

2. I'm always open for suggestions in wording. I'm naturally not good in that, in Portuguese (my native language) , English or German.

3. I think there are some gui ways of adjusting this things, but its complicate to get it right as it depends on screen size. We will have to test a litte, I would just point out this configuration are to be done in the **execute** file

4. You'r right: documentation, documentation, documentation!

Issue #18, opened 2015-05-05
Revision


gamaanderson:
This is just a reminder to handle in the future.

Several parts of the new code were copied from the old one, So I believe we should revise it to make sure it has no bugs hidden, or if it could be done better. I'm specially concerned with the _update_plot() and how the zoom/pan is working.

nguy:
Great idea!  ```python _update_plot``` is a core piece and it is quite possilbe it is not optimized or could be done in a better way. 

Just a comment on Zoom/Pan, I have changed the ToolSelect to no longer be used and have instead added a ```python self.zp``` that will allow a more straighforward way of keeping track here.  It should be noted that when Zoom/Pan is enabled, the arrow keys are blocked.

nguy:
And just to update this thread, The tool setup has changed to now be put into a list.  This was done to hopefully keep track of instances as the functionality grows and makes the checks easier.

@gamaanderson Do you feel better about _update_plot now with the change to V variables (adding strong and weak)?

gamaanderson:
It is an improvement, but I will not fell better till I sit down and understand every line of that code, which I have not done jet. From what I have seen there is a gap between `self.display` and `self.canvas`, I

nguy:
So on line 575 we're using the PyArt library to create figure in matplotlib that is associated to the ```python self.ax```.  This Axis instance was established in the ```python _set_fig_ax``` function, with fig and ax being associated.  The ```python _set_figure_canvas``` function uses the matplotlib backend that connects the matplotlib figure instance and Qt.  So once the figure has been made, L614 is used to actually update the Qt canvas with the matplotlib figure.
Does that make sense?

gamaanderson:
Actually it does, I will reflect over that.

PR #28, opened 2015-05-19
Issue #18

gamaanderson:
This PR is not to be merged, but rather a platform for discussion on specific code topics.


plot.py
```diff
@@ -168,6 +168,7 @@ def LaunchGUI(self):
         # Create the widget
         self.central_widget = QtGui.QWidget()
         self.setCentralWidget(self.central_widget)
+        self._set_figure_canvas()
```
gamaanderson: This just have to executed once, not at every new radar

plot.py
```diff
@@ -399,6 +400,7 @@ def NewRadar(self, variable, value, strong):
         self._fillTiltBox()
         self._fillFieldBox()
 
+        #XXX if: pass, else: pass ?
```
gamaanderson: Is there still a reason for this if, or it can be deleted

plot.py
```diff
@@ -562,12 +564,13 @@ def _update_plot(self):
                                 ax=self.ax, title=self.title)
             self.display.set_limits(xlim=(self.limits['xmin'], self.limits['xmax']),\
                                     ylim=(self.limits['ymin'], self.limits['ymax']),\
-                                    ax=self.ax)
+                                    ax=self.ax, fig=self.fig)
```
gamaanderson: I just believe explicitly give the figure is more readable, even if not needed. The name ax is also not much readable, as it act more like a sub-figure

plot.py
```diff
@@ -562,12 +564,13 @@ def _update_plot(self):
...
             self.display.plot_grid_lines()
         else:
             self.display = pyart.graph.RadarDisplay(self.Vradar.value)
             if self.Vradar.value.scan_type != 'rhi':
                 # Create Plot
+                #XXX this should be handel else where
```
gamaanderson: move this test for newTilt or at the very beginning of _update

plot.py
```diff
@@ -603,6 +606,7 @@ def _update_plot(self):
                                                 norm=norm, orientation='horizontal')
         # colorbar - use specified units or default depending on
         # what has or has not been entered
+        #XXX this should be handel else where
         if self.units is None or self.units == '':
```
gamaanderson: self.Vradar.value.fields[self.field]['units'] should be the standard value for unit not None

plot.py
```diff
@@ -625,6 +629,7 @@ def _check_default_field(self):
         
         This should only occur upon start up with a new file.
         '''
+        #XXX this need re-thinking
```
gamaanderson: Should display be handling this? may be this should be handle in execute?. Or just don't update if field not in radar and wait for the user to change for a valid field

plot.py
```diff
@@ -663,16 +668,19 @@ def _set_default_limits(self):
                 
     def _check_file_type(self):
```
gamaanderson: I just re-write this, I think is aesthetically better this way.

plot.py
```diff
@@ -663,16 +668,19 @@ def _set_default_limits(self):
...
+            if 'platform_type' in radar.metadata:
+                if (radar.metadata['platform_type'] == 'aircraft_tail' or
+                    radar.metadata['platform_type'] == 'aircraft'):
+                    self.airborne = True
+                else:
+                    self.rhi = True
+            else:
                 self.rhi = True
-            
+            #XXX this has only one effect: self._set_default_limits()
```
gamaanderson: if you see this function the only relevant action is self._set_default_limits(). By the way is rhi and airborne working? I just have ppi files.

tilt.py
```diff
@@ -73,8 +73,7 @@ def SetTiltRadioButtons(self):
         self.tiltbutton = []
         
         # Pull out the tilt indices and elevations for display
-        tilt_indices = self.Vradar.value.sweep_start_ray_index['data'][:]
-        elevs = self.Vradar.value.elevation['data'][tilt_indices]
+        elevs = self.Vradar.value.fixed_angle['data'][:]
```
gamaanderson: elevation is the real elevation for every ray, fixed_angle is the nominal target elevation

tools.py
```diff
@@ -105,6 +105,7 @@ def onPick(self, event):
         '''Get value at the point selected by mouse click.'''
         xdata = event.xdata # get event x location
         ydata = event.ydata # get event y location
+        #XXX this fails if not inside plot region
```
gamaanderson: picking outside the plot region generate an error in the terminal, how to avoid this error?

tools.py
```diff
@@ -241,6 +242,7 @@ def onZoom(self, event):
 
         self.ax.set_xlim([xdata - new_width * (1-relx), xdata + new_width * (relx)])
         self.ax.set_ylim([ydata - new_height * (1-rely), ydata + new_height * (rely)])
+        #XXX shouldn't zoom have direct acess to the canvas?
```
gamaanderson: I just find it strange to need to do ax->figure->canvas

plot.py
nguy: I think you are right.  In the original style I believe it was necessary.

nguy: I think this may be fixed with the recent updates?

plot.py
nguy: You took care of this with the reorganization PR.

plot.py
nguy: This is fine, I did correct one instance where it was passed to the wrong routine.

plot.py
nguy: Is this still an issue?

plot.py
nguy: Good idea. I actually ended up pulling the limits initialization into this with the last PR (#36).

plot.py
nguy: Yep, this was poorly coded. I think I just fixed it in PR #36

tilt.py
nguy: I need to look through this a little. When you deal with aircraft radar data, naming conventions can get a little weird (some aircraft radars are rotated 90 degrees from the ground). I don't want to change too much of this, but revisit in the future.

tools.py
nguy: This is fixed now.

tools.py
nguy: I don't follow this one, sorry.

nguy:
Okay, I finally made some comments to these points. Hope they are useful.

Some items may be outdated.

plot.py
gamaanderson: yes, this is fixed

plot.py
gamaanderson: Yes, I do think this is the wrong place for this testing. Also Display should not be overwriting the Vtilt value, its responsibility of the function that is changing Vtilt to assure it has a valid value (0<=value<=len(tilts))

plot.py
gamaanderson: Should we also delete the function `_set_fig_ax_rhi` as it is not being used any more?

nguy: Yes we should. It has no use anymore.

plot.py
nguy: Are you wanting the full path to the units here? I think that it can have the value of None when some file types are read (not containing units or not able to read units).

plot.py
nguy: Well I suppose this could move somewhere else. It is needed when opening a file type for the first time.

plot.py
nguy: It is deleted now in PR #36

plot.py
gamaanderson: We can let this here for now, but I have some ideas I will present later.

PR #33, opened 2015-06-11
Issues 29 and 31

gamaanderson:
This is a big restructuration, may be you should merge it into a separated branch to have to opportunity to give it a good look before accepting. I took opportunity and turn artview into a package and add documentation.

* You can do both: `import artview` and start components directly, or run `python artview -d ... -f ...` in the command line and have it starting in the standard way. Our previous **execute_two.py** is now in **artview/scripts/standard.py**
* There is also a `setup.py`, please revise that to make sure you agree with all points. I've used pyart's `setup.py` as a model, may be we should verify if Jonathan has any objection to that.
* I set up sphinx to build the documentation, give a look into `doc/build/html/index.html`. It still need like a lot of correction, but we have now a structure to work on

### DIFF
--- a/artview/components/menu.py
+++ b/artview/components/menu.py
@@ -102,6 +102,7 @@ class Menu(Component):
         '''Remove widget from central layout.'''
         self.centralLayout.removeWidget(widget)
         self.removeLayoutMenuItem(widget)
+        widget.close()
         widget.deleteLater()
 
     def addComponent(self, Comp):
@@ -211,7 +212,7 @@ class Menu(Component):
     def startComponent(self, Comp):
         '''GUI start a Component and add to layout.'''
         comp = Comp.guiStart()
-        self.addLayoutWidget(comp)
+        #self.addLayoutWidget(comp)
 
     def addFileAdvanceMenu(self):
         '''Add an option to advance to next or previous file.'''
@@ -362,6 +363,7 @@ class Menu(Component):
         # Read the data from file
         try:
             radar = pyart.io.read(self.filename, delay_field_loading=True)
+            radar.fields['PHIDP']['data'] = np.fmod(radar.fields['PHIDP']['data']+180,360)
             self.Vradar.change(radar)
         except:
             msg = "This is not a recognized radar file!"
--- a/artview/components/plot.py
+++ b/artview/components/plot.py
@@ -17,7 +17,7 @@ from matplotlib.pyplot import cm
 
 
 
-from ..core import Variable, Component, common
+from ..core import Variable, Component, common, VariableChoose
 
 # Save image file type and DPI (resolution)
 IMAGE_EXT = 'png'
@@ -33,6 +33,11 @@ class Display(Component):
     Class that creates a display plot, using a returned Radar structure 
     from the PyArt pyart.graph package.
     '''
+
+    @classmethod
+    def guiStart(self, parent=None):
+        args = _DisplayStart().startDisplay()
+        return self(**args)
 
     def __init__(self, Vradar, Vfield, Vtilt, Vlims=None, \
                  airborne=False, rhi=False, name="Display", parent=None):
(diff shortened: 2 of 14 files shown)


nguy:
@gamaanderson  I just branched this to start digging into the code. Given the level of changes and me being in the field on a project currently, it will take me a bit to get through them likely. But I'll comment as I go.

artview/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+from . import plugins
+from . import scripts 
+
+
+run = scripts.scripts['standard']
+
+
+if __name__ == "__main__":
+    import sys
+    print "haha"
```
nguy: This is a neat way to handle this. I guess that we could also have different variants.
Something we should consider is a "light-weight" version in the future.

artview/components/menu.py
```diff
@@ -361,7 +361,7 @@ def _openfile(self):
      
         # Read the data from file
         try:
-            radar = pyart.io.read(self.filename)
+            radar = pyart.io.read(self.filename, delay_field_loading=True)
```
nguy: We need to make sure we are smart in how we handle errors here. Some file types do not support the 'delay_field_loading' parameter. Does PyART handle this internally and just return all fields if it is not used?

artview/components/plot.py
```diff
@@ -61,13 +61,14 @@ def __init__(self, Vradar, Vfield, Vtilt, Vlims=None, \
...
             If None, then Qt owns, otherwise associated with parent PyQt instance.
-        
+
         Notes::
         -----
         This class records the selected button and passes the 
         change value back to variable.
         '''
         super(Display, self).__init__(name=name, parent=parent)
+        self.setFocusPolicy(QtCore.Qt.ClickFocus)
```
nguy: What does this line do (that is different than without)?

artview/components/plot.py
```diff
@@ -381,7 +382,6 @@ def NewRadar(self, variable, value, strong):
         '''Display changes after radar Variable class is altered.'''
         # In case the flags were not used at startup
         self._check_file_type()
-        self._set_figure_canvas()
```
nguy: So did you find _set_figure_canvas() not needed?

artview/components/tilt.py
```diff
@@ -72,8 +72,7 @@ def SetTiltRadioButtons(self):
         self.tiltbutton = []
         
         # Pull out the tilt indices and elevations for display
-        tilt_indices = self.Vradar.value.sweep_start_ray_index['data'][:]
-        elevs = self.Vradar.value.elevation['data'][tilt_indices]
+        elevs = self.Vradar.value.fixed_angle['data'][:]
```
nguy: Nice catch on this one to remove the tilt_indices variable!

artview/components/tools.py
```diff
@@ -105,6 +105,7 @@ def onPick(self, event):
         '''Get value at the point selected by mouse click.'''
         xdata = event.xdata # get event x location
         ydata = event.ydata # get event y location
+        #XXX this fails if not inside plot region
```
nguy: Does this refer to the error output to terminal?

artview/parser.py
```diff
@@ -7,10 +7,10 @@
 import argparse
 
 # Get the version
-f = open("version", "r")
-NAME = f.readline().strip()
-VERSION = f.readline().strip()
-f.close()
+import version
```
nguy: I don't see a version.py in the distribution. Was this supposed to incorporate the old script or is this a separate package?

nguy:
@gamaanderson I've made some comments throughout. Please check.

@jjhelmus We've (actually @gamaanderson ) made ARTView an installable program now. Any objections of us relying heavily on Py-ART setup and structure? The ideas was to maintain some continuity for future users/developers.

setup.py
```diff
@@ -0,0 +1,190 @@
+#!/usr/bin/env python
+"""ARTview - the ARM Radar Toolkit Viewer.
+
+The ARM Radar Toolkit Viewer, ARTview, is an open source Python module that
+give acess to Py-ART funtion within a PyQt graphical interface
+"""
```
nguy: This line """ needs to be moved to L18.

artview/__init__.py
gamaanderson: Possibilities are all there, is just a question of implementing

artview/components/menu.py
gamaanderson: definitely this is a PyART bug, `pyart.io.read` is just getting **kwargs and passing it to the other read functions, no test for `delay_field_loading` is made. I will work with @jjhelmus to correct this.

gamaanderson: while that, let the original version

artview/components/plot.py
gamaanderson: This is probably system dependent, but at one point I could not use zoom because the mouse wheel was changing focus. So this what the solution I found. I believe the standard configuration is [Qt.WheelFocus](http://pyqt.sourceforge.net/Docs/PyQt4/qt.html#FocusPolicy-enum)

artview/components/plot.py
gamaanderson: just in this spot, still need in `LaunchGUI`

artview/components/tools.py
gamaanderson: yes it does, at least for me. I needed some time to track down the reason, but I couldn’t come up with a simple solution, so I just marked it.

artview/parser.py
gamaanderson: this will be created while installing, for instance in place with `python setup.py build_ext -i`. We need to add some install instructions

setup.py
gamaanderson: yes it does!

gamaanderson:
Just pushed some bug corrections:

* undo the use of `delay_field_loading` til we get it stable
* correct some imports to the correct local form
* remove raw text from setup.py
* add rudimentary install instruction in  INSTALL.rst

artview/components/plot.py
nguy: Okay, I wondered because I never experienced that on my machine. We may run across more as users try this package out.

artview/components/plot.py
nguy: gotcha

artview/components/tools.py
nguy: Agreed, it's one I considered minor for now, but you're right we should take care of it.

nguy:
@gamaanderson I assume you were able to install this? 
I am not able, mostly due to the relative import links. Outside of the ```python python -m ``` install is there another way that I am missing?

gamaanderson:
I've forgotten one file: **artview/setup.py**

installation is done like pyart:
`python setup.py install`
or
`python setup.py install --user`

if you are running artview as a script outside the main folder you need
`python -m artview`

In the process of pushing this file I accidentally push some other alteration (plugins), that I would rather have pushed afterwards. Any way if you want you may ignore them for now, or if you want explore the `components` menu in the menubar. I will bring up a discussion about those additions later.

jjhelmus:
Very cool.  You should be able to depend on the public API (all function, classes and modules in the User documentation which are not prefixed by an underscore) of Py-ART staying stable.  If I do break anything let me know and I can either revert it or provide a work around.  I'll fix the delay field loading issue in Py-ART shortly.

nguy:
@gamaanderson - That last fix made all the difference!
After I accept the changes, I'll make a pass through to things like update Readme and link to Install and Documentation. Per the documentation, where does once I push the changes to the master branch?

I want to spend just a little more time digging in to familiarize myself with the changes. I noticed the Components underwent a really big change. I am trying to grasp the flow here. Maybe we should start an Issue to keep track of development?

nguy:
@jjhelmus - Thanks for the input. We'll let you know if things don't play nice, but will try to fix whatever we can on our end.

gamaanderson:
@nguy open issues as you want, there is in fact a lot to be changed, and that is deliberate, I decided to first change the structure to them change the content. Once you have merged and point out points of concern I can get to work, otherwise I try to not be in the way. 

The plugins I've added are just front end for executing pyart.corrections. While doing that I was able to fix an old request from you, a GUI start for Display. 

sorry, but I didn't understand the following sentence: "Per the documentation, where does once I push the changes to the master branch?"

nguy:
Thanks @gamaanderson I'll probably open an issue later on so we can keep a conversation present on Component development.

Hopefully I'll merge this later today. I have to get out to the airplane for a bit of work now.

I was trying to figure out where the documentation exists when created. Probably need to look on Github to figure this out.

gamaanderson:
Github don't create documentation by its own, we must create it and place in a separated branch. Unfortunately that don't combine easily with sphinx, so I didn't set that. But you can open in your computer with the browser the file ` artview/doc/build/html/index.html` and have a look on it

nguy:
I see it now. Probably better to work on the documentation some before release. And to get some of the bugs worked out first. I'm going to merge now and then make a couple of small changes

nguy:
@gamaanderson I know you told me where the documentation from my branch would be located, but I can't find it anywhere.
Thinking of updating the readme with next pass.
I'll add documentation link, workarounds for old versions (that we know) and versions used for testing thus far.

gamaanderson:
here it is:
https://rawgit.com/nguy/artview/master/docs/build/html/index.html

nguy:
Awesome, thanks!

PR #34, opened 2015-06-18
ROI works and creates a output file; minor fixes to onPick

pfhein:
You can now select a region of interest and the points in the region.  A table of points is displayed in the terminal window and also is saved to a file.  Three files were changed: plot.py, artview.py, and tools.py.  Most of the changes were in tools.py in ROI.

Also in onPick I found that the right point was not always being found between 359 and 0 degree azimuths.  I figured out how to fix it.


gamaanderson:
I'm looking forward  to get a look on this. I would suggest to merge this PR before PR #33, this will probably track better files that were moved. Also as it appears sync was done manually, its probably better to just merge files with actual alterations to avoid conflicts.

pfhein:
Yes, I don't know the proper way to sync so commits appear be a mess.  
It is the current version (at the end of May) with modifications to 
those 3 files, mainly to tools.py.  You can just diff the files and see 
what needs to be changed.

nguy:
I'll have to look at the code, but since I've started down the path with PR #33 I may institute this as a tool afterward. I'll comment again once I get a bit further.

nguy:
I have this partially implemented - I have the code put into the tools.py infrastructure. But what I think would be most useful would be to display this as a QTableWidget. I haven't quite figured this out yet but am working on this.

nguy:
Another task will be to clear the current ROI as you draw another (or let user clear?)

pfhein:
You can double click on the tools button and that does clear the the 
current ROI.  Not the most elegant way to do it, but it works.

nguy:
Okay, I implemented this code with a few tweaks. It calls a Dialog box that opens a tabular display of values, can save a .csv file and can read a .csv file, and resets the display (i.e. clears the selected region). I foresee the opening a file being useful in the future, but not really right now. I changed the format of the output file because a csv is relatively easy to deal with in python (and I found nice coe online!). Check out pull request #36. If you see improvements have at it.

nguy:
This PR was rolled into PR #36

PR #36, opened 2015-06-25
Updated documentation, ROI, and file reading.

nguy:
Instituted a modified version of the Region of Interest PR #34.
Added save, open, and reset capabilities of ROI selection.
In addition added a Table output window for ROI selection.
Updated the restore_to_default in tools.py  to be more robust as more tools are added.
RHI and airborne file types are now working once again.


gamaanderson:
This looks good, for me it ok merging it.

artview/components/plot.py
```diff
@@ -112,23 +103,26 @@ def __init__(self, Vradar, Vfield, Vtilt, Vlims=None, \
...
-        # Create a figure for output
-        self._set_fig_ax(nrows=1, ncols=1)
-
         # Create tool dictionary
         self.tools = {}
-        self.tools['zoompan'] = None
+
+        # Check the file type and initialize limts
+        # Create a figure for output
+        self._check_file_type()
```
gamaanderson: don't need _check_file_type here, newRadar does it and will be executed in line 126

nguy: I'll need to look at this, but I know the _set_fig_ax function needs _check_file_type to be run first to get the proper limits set. In turn the LaunchGUI requires the all of this to be done so that the "canvas" can be set up properly. 

That is why it has persisted in two locations. Maybe there is an elegant way to treat this problem, but I haven't figured it out yet.

gamaanderson: Ok, so there is a conflict between initialization and execution, this indicates we are doing something wrong.
From the start: when should the limits be reset?

nguy: The _initialize limits need to be set in three circumstances:
1. before drawing the initial plot 
2. when the plot type (i.e. ppi goes to rhi) changes
3. when the reset default is chosen.

This causes a "hard" reset to a default state.
Maybe this could be broken into a couple of different functions. The reason the figure needs it is to define the aspect ratio of the plot (RHI and Airborne both have longer X than Y axes). This was convenient a while back, but could probably be changed now.

nguy:
@gamaanderson I renamed plot.py to plot_radar.py and made some structural changes per your line of thought. Basically I create figure first and then modify it later from withing the check_file_type. This allows this operation to be performed with NewRadar, taking care of all three instances I believe.

artview/components/plot_radar.py
```diff
@@ -648,7 +605,7 @@ def _check_file_type(self):
         '''Check file to see if the file is airborne or rhi.'''
         radar = self.Vradar.value
         if radar.scan_type != 'rhi':
-            return
+            pass
```
gamaanderson: reset flags

    self.rhi = False
    self.airborne = False

artview/components/plot_radar.py
```diff
@@ -658,8 +615,8 @@ def _check_file_type(self):
                     self.rhi = True
             else:
                 self.rhi = True
-            #XXX this has only one effect: self._set_default_limits()
-            self._set_fig_ax_rhi()
+        self._set_default_limits()
```
gamaanderson: this is reseting limits at every new file, not just when file-type changes

artview/components/plot_radar.py
```diff
@@ -658,8 +615,8 @@ def _check_file_type(self):
...
+        self._set_fig_ax()
```
gamaanderson: open a PPI file and them a RHI, I believe the re-execution of _set_fig_ax is causing the problem

artview/components/plot_radar.py
nguy: Good to be explicit here.

artview/components/plot_radar.py
nguy: You are correct.

Issue #32, opened 2015-05-20
Grid and map


gamaanderson:
@nguy I just opened some new issues to lead us in a new phase of the development. Now, however, I'm here as an user, I were in need of the following two functionalities: a display for grid data and execute pyart.map.grid_from_radars. The intent is to calibrate the interpolations option to get the best equilibrium between quality and execution time. 

If we were able to develop this in the next month or so would be great. I know you have other projects and so do I, so handle things as you got time.

nguy:
Interesting.  Yeah I'll work on these as time permits.  My next week and half is very busy with aircraft deployment.  When in the field, I may have down time (depending on weather) to work on these issues.

My first instinct would be to revisit if PyArt spits out any kind of message (or attribute) upon IO.  Let me put some thought into this.

gamaanderson:
I'm adding here a list o task to close this issue, the first two are the most important, the other are just accessory

- [ ] Add plot_grid.py similar to plot.py, with variable Vgrid
- [ ] Add component (plugin) with two variable Vradar, Vgrid. It execute grid_from_radars from Vradar and put the result in Vgrid. It have a form to fill up the interpolation options and a map button that start the mapping.
- [ ] Make Menu work with radars and grids
- [ ] Make TiltButtonWindow work with grid as well
- [ ] Rename plot.py to plot_radar.py or something like that

suggest changes to this list as you wish

nguy:
Sorry I'm just getting to this.

I forgot to do this in the latest PR, but I'd like to rename plot.py to plot_radial.py in an upcoming PR.

As for the grid, would it work to add a way to access the ```python grid_from_radar``` to the tools or plugins? Or do you have gridded files athat need read in as well?

gamaanderson:
Yes, reading grid files is also important. I don't think we should make Display work with grid, but rather have a GridDisplay.

nguy:
Okay, so you want to do the gridding in an ARTView function?
Versus using a [PyART](http://arm-doe.github.io/pyart-docs-travis/user_reference/generated/pyart.graph.GridMapDisplay.html#pyart.graph.GridMapDisplay) method?

gamaanderson:
no, still using PyART. I just don't want the Display class working with grids, it needs a different class in a different file, that probably start as a copy of Display. I already have this in a branch, I will make a pull request so that you can give a look.

nguy:
Sounds good.

gamaanderson:
[...]

PR #37, opened 2015-06-25
Issue #32

gamaanderson:
add preliminary support for grid and mapping.
File artview/components/plot_grid.py is outdated to recent improvements of plot.py and also need intensive cleaning up


nguy:
@gamaanderson Any chance you could pull over the newest push and see what you can merge on your side? I hate to make more work for you. If you need, I can see about putting some of this into the master and we can do this change in increments.

gamaanderson:
No problem, I agree this is the best way to proceed.

PR #46, opened 2015-07-02
Roi component

gamaanderson:
This is a suggestion for a change in ROI and I want to emphasize as a suggestion this can also be rejected, I can also see reasons for that. This is mostly personal preference and a sense of what is better that I can't really explain, there is however one positive point, the new ROI is independent of Vradar and therefore can be used (in the future) with grid.

The new ROI may be found in the components menu in the menubar. But this is open to discussion.

Take your time with this, this is NOT a priority and I pretend to focus more in the documentation

### DIFF
--- a/artview/components/plot_radar.py
+++ b/artview/components/plot_radar.py
@@ -466,7 +466,7 @@ class Display(Component):
                                           self.scan_type, self.rhi)
         self._update_plot()
 
-    def pathInteriorValues(self, path):
+    def getPathInteriorValues(self, path):
         '''
         Return the bins values path.
 
@@ -680,6 +680,16 @@ class Display(Component):
         if path:
             self.canvas.print_figure(path, dpi=DPI)
             self.statusbar.showMessage('Saved to %s' % path)
+
+    ########################
+    #      get methods     #
+    ########################
+
+    def getPlotAxis(self):
+        return self.ax
+
+    def getStatusBar(self):
+        return self.statusbar
 
 
 class _DisplayStart(QtGui.QDialog):
--- a/artview/components/plot_radar.py
+++ b/artview/components/plot_radar.py
@@ -466,6 +466,30 @@ class Display(Component):
                                           self.scan_type, self.rhi)
         self._update_plot()
 
+    def pathInteriorValues(self, path):
+        '''
+        Return the bins values path.
+
+        Parameters
+        ----------
+        path - Matplotlib Path instance
+
+        Returns
+        -------
+        x, y, azi, range, value, ray_idx, range_inx: ndarray
+            Truplet of 1arrays containing x,y coordinate, azimuth,
+            range, current field value, ray index and range index
+            for all bin of the current radar and tilt inside path
+        '''
+        from .tools import interior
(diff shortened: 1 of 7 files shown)


nguy:
I like the idea behind this and was thinking of something similar in terms of relocating the ROI function as a component. I'll need to look through and see how this is set up. 

By the way, is there documentation of how to call a component? I think this will be important to let users know how to use the plugins effectively.

gamaanderson:
I'm working in a tutorial for writing scripts right now, this may help, but keep this idea we will need it

nguy:
Good deal, thanks!

I've been doing some mental coding (trying to get a structure in my head) with some of these issues. I am trying to decide what would be really nice to have done by Sep AMS radar conference. We'll be presenting ARTView as an option in the open source software short course. So any tutorials you think might be useful please let me know. 

Hopefully I can get back to this soon. I have a couple other things I need to finish up.

gamaanderson:
You can keep up with what I'm doing [here](https://rawgit.com/gamaanderson/artview/documentation/docs/build/html/index.html). The reference manual is a lot more complete, the tutorial is a work in progress. Just let me say I'm famous for some terrible spelling errors, there is probably a lot there.

If you have clear ideas about what you need, let me know.

artview/components/plot_radar.py
```diff
@@ -466,6 +466,30 @@ def toolDefaultCmd(self):
...
+        '''
+        from .tools import interior
+        radar = self.Vradar.value
+        xy, idx = interior(path, self.Vradar.value, self.Vtilt.value)
+        aux = (xy[:, 0], xy[:, 1], radar.azimuth['data'][idx[:, 0]],
+               radar.range['data'][idx[:, 1]] / 1000.,
+               radar.fields[self.Vfield.value]['data'][idx[:, 0], idx[:, 1]],
+               idx[:, 0], idx[:, 1])
+        return aux
+
```
nguy: I think this is a really good idea. This routine should become useful with future development.

artview/components/plot_radar.py
```diff
@@ -657,6 +681,16 @@ def _savefile(self, PTYPE=IMAGE_EXT):
...
+    ########################
+    #      get methods     #
+    ########################
+
+    def getPlotAxis(self):
+        return self.ax
+
+    def getStatusBar(self):
+        return self.statusbar
+
```
nguy: Nice, these are much more elegant solutions!

artview/components/plot_radar.py
gamaanderson: The base here is that ROI is about space, not the data that is in that space, so it should not have access to Vradar. As a consequence Display must provide this information.

nguy:
Okay, so this all seems to work just fine. 
I guess I don't really know how the components work. I see that ROI is now a component as well as tool (let's keep it that way for now). But selecting it does not "run" anything.

Could you note the procedure here or elsewhere (maybe you've done this already and I don't recall)?

gamaanderson:
Ok, that is missing a warning, you must select a display before starting it. You probably just clicked start right way and that resulted in an error in the terminal

PR #48, opened 2015-07-09
DOC: tutorial

gamaanderson:
Ok, this is a lot of documentation addiction, so much that I wouldn't try to see the alterations, just browser in the html: https://rawgit.com/gamaanderson/artview/documentation/docs/build/html/index.html
You may found some interesting things (including hyperlinks to pyart and mathplotlib documentation, PyQt unfortunately does not allow that)

I'm stepping down from documentation for now, that does not mean it is ready, there are still errors, I can't see them but I know they are there.


nguy:
Is there a good way to merge multiple pull requests. Seems like accepting one always makes the other unable to merge?

nguy:
This was merged manually.

PR #49, opened 2015-07-10
Changes

gamaanderson:
changes include:
Make None a valid value for Vradar, this is basic adding some tests.
Add print traceback if pyart read fails
Handel fail in selecting file
correct bug in Display._open_tiltbuttonwindow
add warning in roi.py for selecting display
turn common.ShowLongText into modal or modalles
correct CreateTable.display_data case no data

### DIFF
--- a/artview/components/menu.py
+++ b/artview/components/menu.py
@@ -102,10 +102,14 @@ class Menu(Component):
     def showFileDialog(self):
         '''Open a dialog box to choose file.'''
 
-        self.qfilename = QtGui.QFileDialog.getOpenFileName(self, 'Open file', 
+        filename = QtGui.QFileDialog.getOpenFileName(self, 'Open file', 
                 self.dirIn)
-        self.filename = str(self.qfilename)
-        self._openfile()
+        filename = str(filename)
+        if filename == '':
+            return
+        else:
+            self.filename = filename
+            self._openfile()
 
     def addLayoutWidget(self, widget):
         '''
@@ -264,14 +268,15 @@ class Menu(Component):
         txOut = "This is a simple radar file browser to allow \
                  quicklooks using the DoE PyArt software"
         QtGui.QMessageBox.about(self, "About ARTView", txOut)
- 
+
     def _get_RadarLongInfo(self):
-        '''Print out the radar info to text box.'''
+        '''Print out the radar info to text box and terminal.'''
         # Get the radar info form rada object and print it
         txOut = self.Vradar.value.info()
-        print txOut
+        common.ShowLongText(txOut, modal=False)
         
-        QtGui.QMessageBox.information(self, "Long Radar Info", "See terminal window") 
+        print txOut
+#        QtGui.QMessageBox.information(self, "Long Radar Info", "See terminal window") 
 
     def _get_RadarShortInfo(self):
         '''Print out some basic info about the radar.'''
@@ -331,7 +336,7 @@ class Menu(Component):
             nsweeps = str(self.Vradar.value.nsweeps)
         except:
             nsweeps = "Info not available"
-        
+
         txOut = (('Radar Name: %s\n'% rname) +\
         ('Radar longitude: %s\n'% rlon) + \
         ('Radar latitude: %s\n'% rlat) + \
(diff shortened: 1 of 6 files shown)


artview/components/menu.py
```diff
@@ -264,13 +268,13 @@ def _about(self):
         txOut = "This is a simple radar file browser to allow \
                  quicklooks using the DoE PyArt software"
         QtGui.QMessageBox.about(self, "About ARTView", txOut)
- 
+
     def _get_RadarLongInfo(self):
-        '''Print out the radar info to text box.'''
+        '''Print out the radar info to text box and terminal.'''
```
nguy: I forgot that I had not found a way to print this to a text window yet.
Just made a note to look into this.

gamaanderson: actually we already have a way of displaying long texts [here](https://github.com/nguy/artview/blob/master/artview/core/common.py#L53), I'm using this for the help button in the plug-ins.
The problem is that txOut is None, `.info()` is doing the print, not `print txOut`, see the [pyart documentation](http://arm-doe.github.io/pyart-docs-travis/user_reference/generated/pyart.core.Radar.info.html)
this can only be solved with a [string buffer](https://docs.python.org/2/library/stringio.html)

artview/components/tools.py
```diff
@@ -110,6 +110,13 @@ def onPick(self, event):
         ydata = event.ydata # get event y location
         if (xdata is None) or (ydata is None):
             self.msg = "Please choose point inside plot area"
+        elif self.Vradar.value is None:
+            az = np.arctan2(xdata, ydata)*180./np.pi
+            if az < 0:
+                az = az + 360.
+            rng = np.sqrt(xdata*xdata + ydata*ydata)
+            self.msg = 'x = %4.2f, y = %4.2f, Azimuth = %4.2f deg., Range = %4.3f km'\
+                        %(xdata, ydata, az, rng)
```
nguy: If Vradar is None, should the calculation be performed?
Not sure how this is different from the else portion.

artview/core/common.py
```diff
@@ -50,7 +50,7 @@ def ShowQuestion(msg):
     return response
 
 
-def ShowLongText(msg):
+def ShowLongText(msg, modal=True):
```
nguy: Can you explain modal to me?

artview/core/common.py
```diff
@@ -160,9 +164,12 @@ def display_data(self, data):
             the table """
 
         if len(data)==0:
-            data = ["No data for selected ROI."]
+            data = [["No data for selected ROI."]]
+            nrows = 0
+            ncols = 0
+        else:
+            nrows, ncols = data.shape[0], data.shape[1]
```
nguy: Does this occur if you select outside of plot region?
If you select inside plot region, there is still data, correct?

artview/components/menu.py
nguy: I have to admit this is REALLY low on my to-do. So it's on my list but I won't get to this anytime soon.

artview/components/tools.py
gamaanderson: if radar is none, there is no data, but x,y still exist and azi and range as polar coordinates also do so I decided to print that instead of just giving a warning. In else you will get the azi, rng from the actual bean and the value.

artview/core/common.py
gamaanderson: modal dialog (executed with `exec_()`) lock in the position and just continue the code when closed, other windows may also lock/ get faded etc.
Modales dialog (executed with `show()`) continue execution normally and the window can stay open without affecting the others.

artview/core/common.py
gamaanderson: 1. Yes (if you mean outside the radar region: the circle region not the square window)
2. yes if radar is not None, otherwise data will be empty

artview/components/tools.py
nguy: That works.

artview/core/common.py
nguy: Okay, great.

nguy:
@gamaanderson Note that I added a couple of minor changes to fix airborne plotting crash and README.

PR #52, opened 2015-07-13
More pep8

tjlang:
By request …


nguy:
Thanks @tjlang 
I also switched the PyArt colormaps so make sure to get the latest!

PR #55, opened 2015-07-17
Complet PR #49

gamaanderson:
I mixed up the branches and some alteration didn't get into PR #49. Here are they.


artview/components/tools.py
```diff
@@ -122,16 +123,18 @@ def onPick(self, event):
...
-            radar = self.Vradar.value  # keep equations clean
+            az = np.arctan2(xdata, ydata)*180./np.pi
+            radar = self.Vradar.value #keep equantions clean
             if az < 0:
                 az = az + 360.
             rng = np.sqrt(xdata*xdata + ydata*ydata)
+            # TJL - Attempt to pep8 this overlong string
+            msg1 = 'x = %4.2f, y = %4.2f, ' % (xdata, ydata)
+            msg2 = 'Azimuth = %4.2f deg., Range = %4.3f km' % (az, rng)
+            self.msg = msg1 + msg2
```
nguy: Why is this different that above?

Also, can we keep spaces on mathematical operators?

gamaanderson: I didn't change this, I has something to do with @tjlang pep8 PR.

Not only one can, but it is advised to keep space on all operators

nguy: Gotcha. At some point we'll run the pep8 again and see what happens.

Issue #43, opened 2015-06-29
Pyart Config


gamaanderson:
We are currently dealing with some ugly hacks, for instance [this one](https://github.com/nguy/artview/blob/5c230a218cbe3439d755086bfb38b2f1725e2006/artview/components/plot_radar.py#L580-L607), that I believe the correct way of dealing would be using pyart configuration tools with the follow principle: "artview should always work with a unconfigured pyart, but a more smooth user experience does presuppose pyart is correctly configured"

However, for that to work we need pyart to give more focus to this possibility, this could be as simple as mentioning the PYART_CONFIG flag in README. What do you think of it @jjhelmus?

Related to this, we should also configure standard values for the scale (vmin and vmax) and colormap for every (pyart-)field. We are currently using values that are only useful in reflectivity-like fields, I have dificulties even for visualizing radial velocity. Couldn't pyart.graph also benefit from those standards?

jjhelmus:
@gamaanderson I do not think there is a elegant solution to the multiple field name issue, but I would be open to examining any that you have in mine.  The variety of moment names and conventions has been a pain point in Py-ART that the configuration logic and PYART_CONFIG environment variable attempts to address.  One possibility would be to have ARTView use the `pyart.config.get_field_name` to retrieve the field names which are set in the default or user defined configuration file.  This is what Py-ART uses to generate field names within correction and other functions if they are not explicitly specified.  This would not solve the issue at hand with the multiple names but would at least support users who customization their configuration in Py-ART.

+1 on adding a note about PYART_CONFIG to the README, I'll create a Py-ART issue for this.  

Also +1 on working out a method for standardizing the scale and colormap which pyarty.graph and ARTView can use.  Having these customizable in the same manner as the field names would make sense to me.  There will be some challenges in standardizing these as "good" values often depend significantly on the radar hardware and settings.  For example radial velocities should have limits of +/- the Nyquist velocity which changes depending on the radar and in some cases even the sweep.  Codifying this into a function may be challenging.  I'll make an issue in Py-ART.

nguy:
@gamaanderson nearly all fields do have a vmin and vmax associated with them in ```components/limits.py``` in the ```_default_limits```. These should be pretty standard, because once the reflectivity field is properly read in (which is why the hack exists) all other fields are mapped accordingly (as far as I've seen - maybe this isn't true?).

[...]

PR #56, opened 2015-07-17
Limits

gamaanderson:
Here I did two things:

1. change `_default_limits` to pyart names as a step to implement issue #43 

2. change how display handle limits and colormap: remove limits (dict) and CMAP (string) and added Vlims and Vcmap both Variables holding dictionary.
   *  Vlims keys: 'xmin' (float), 'xmax' (float), 'ymin' (float), 'ymax' (float)
   *  Vcmap keys: 'vmin' (float), 'vmax' (float), 'cmap' (string)

  this end up resulting in change throughout code, but while doing I really had the impression this way is much better, for instance as `ZoomPan` now work over `Vlims` this allow it to affect several display at the same time (just connect Vlims in ComponentControl)

### DIFF
--- a/artview/components/limits.py
+++ b/artview/components/limits.py
@@ -6,7 +6,7 @@ Routines used for modifying limits via Display window.
 
 # Load the needed packages
 from PyQt4 import QtGui, QtCore
-
+import pyart
 
 def _default_limits(field, scan_type):
     '''
@@ -30,11 +30,13 @@ def _default_limits(field, scan_type):
     VR_LIMS = (-30., 30.)
     ZDR_LIMS = (-5., 5.)
     RHO_HV_LIMS = (.8, 1.)
-    KDP_LIMS = (0., 5.)
-    PHIDP_LIMS = (0., 1.)
+    KDP_LIMS = (-2., 5.)
+    PHIDP_LIMS = (-180, 180)
     NCP_LIMS = (0., 1.)
     SW_LIMS = (-1., 10.)
     TP_LIMS = (-200., 100.)
+    SNR_LIMS = (0., 65.)
+    RR_LIMS = (0., 150.)#Rrait12
 
     # X, Y range and size for airborne file typesAIR_XRNG = (-150., 150.)
     AIR_XRNG = (-50., 50.)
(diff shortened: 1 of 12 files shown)


artview/components/plot_radar.py
```diff
@@ -83,10 +87,16 @@ def __init__(self, Vradar, Vfield, Vtilt, Vlims=None,
...
+            self.Vcmap = Variable(None)
+        else:
+            self.Vcmap = Vcmap
+
         self.sharedVariables = {"Vradar": self.NewRadar,
                                 "Vfield": self.NewField,
                                 "Vtilt": self.NewTilt,
-                                "Vlims": self.NewLims}
+                                "Vlims": self.NewLims,
+                                "Vcmap": self.NewCmap,}
```
nguy: So this Vcmap is now a shareable variable correct? This way the information could be ported between various windows?

artview/components/limits.py
```diff
@@ -208,11 +209,11 @@ def limits_dialog(limits, name):
...
-        limits['vmax'] = float(ent_dmax.text())
+        cmap['vmin'] = float(ent_dmin.text())
+        cmap['vmax'] = float(ent_dmax.text())
         limits['xmin'] = float(ent_xmin.text())
         limits['xmax'] = float(ent_xmax.text())
         limits['ymin'] = float(ent_ymin.text())
         limits['ymax'] = float(ent_ymax.text())
 
-    return limits, retval
+    return limits, cmap, retval
```
nguy: The reason for this change is due to adding Vcmap?

artview/components/plot_radar.py
```diff
@@ -104,14 +114,18 @@ def __init__(self, Vradar, Vfield, Vtilt, Vlims=None,
...
         # Create tool dictionary
         self.tools = {}
 
-        # Set up Default limits
-        self._set_default_limits()
+        # Set up Default limits and cmap
+        if Vlims is None:
+            self._set_default_limits(strong=False)
+        if Vcmap is None:
+            self._set_default_cmap(strong=False)
```
nguy: Aha, now I think I am seeing the versatility of the shared limits and cmap variable.

artview/components/plot_radar.py
```diff
@@ -211,22 +222,6 @@ def _fillFieldBox(self):
...
-        if entry['xmin'] is not None:
-            self.limits['xmin'] = entry['xmin']
-        if entry['xmax'] is not None:
-            self.limits['xmax'] = entry['xmax']
-        if entry['ymin'] is not None:
-            self.limits['ymin'] = entry['ymin']
-        if entry['ymax'] is not None:
-            self.limits['ymax'] = entry['ymax']
-        self._update_plot()
-
```
nguy: So this goes away as the change is recorded in Vlims now, right?

artview/components/plot_radar.py
gamaanderson: that is correct.

moreover, this could also be used for instance in a plugin that draw a more sophisticated colorbar

artview/components/plot_radar.py
```diff
@@ -388,16 +385,18 @@ def NewRadar(self, variable, value, strong):
...
+        * Reset colormap
         * Reset units
         * Update fields MenuBox
         * If strong update: update plot
         '''
-        self._set_default_limits()
+        self._set_default_cmap(strong=False)
         self.units = None
         idx = self.fieldBox.findText(value)
         self.fieldBox.setCurrentIndex(idx)
```
nguy: Why do the limits not change according to the new field?

artview/components/limits.py
gamaanderson: this is because vmin and vmax are no longer part of limits, but rather of Vcmap. The user however don't see the difference (may be he should, may be he should not, I just didn't change that)

artview/components/plot_radar.py
gamaanderson: this function was already not being used

artview/components/plot_radar.py
gamaanderson: because limits is basic the zoom, it doesn't have to change, just the colorbar (with vmin and vmax) changes

artview/components/limits.py
nguy: I haven't looked yet, but is the scaling option still there for the vmin, vmax?

artview/components/plot_radar.py
nguy: Ha, didn't even realize that!

nguy:
A general comment, does the potential change in PyART issue [#329](https://github.com/ARM-DOE/pyart/issues/329) have an affect on this?  Should we wait for changes if so?

artview/components/limits.py
gamaanderson: yes, the graphical part is all the same

gamaanderson:
A small re-visitation may be needed when that issue is handle, however I see no reason to wait.

gamaanderson:
I mean, that issue is an enhancement, it won't break anything.

artview/components/plot_radar.py
```diff
@@ -513,10 +534,10 @@ def toolCustomCmd(self):
     def toolDefaultCmd(self):
         '''Restore the Display defaults.'''
         from . import tools
-        self.tools, self.limits, self.CMAP = tools.restore_default_display(
+        self.tools, limits, cmap = tools.restore_default_display(
```
nguy: I think that the cmap instance should be removed here, correct?

gamaanderson: no, the problems here are much bigger, I'm making a commit to correct it

gamaanderson: done, it wasn't actually that difficult.

gamaanderson:
Thank you for merging the last alterations, I will use the current version of plot_radar to create plot_grid. I will try to avoid code duplication, but it will mostly probably have some

nguy:
Sounds like a plan. 

Just and FYI, I'm going to be working on getting some the ```exclude``` functions into ARTView over the next few weeks (instead of the mask I proposed in PyART issue). I haven't fully formed how I want to do this yet. It shouldn't affect any changes you make I think.

gamaanderson:
Ok, but please keep me informed as you progress on that. Also an FYI, see this [commit](https://github.com/gamaanderson/pyart/commit/5f393aac28eab50330350fd94aa3d40cd709aa3c) of mine, this will give a more direct access to GateFilter so we can for instance built a filter from roi. I haven't open a pyart PR yet, just because I like to go by steps

nguy:
Will do @gamaanderson that commit looks great - it will be very useful for ARTView.

I need to turn my focus on finishing a paper for the next couple of weeks, so expect some more work after that. In August, I'll ping the group for any last minute additions before the radar conference.

PR #58, opened 2015-07-27
Grid + Nomeclature change

gamaanderson:
I did something that may be I shouldn't: to match the updates in grid display I changes the nomenclature of other parts of the code.

* changed `Display` to `RadarDisplay` and `GridDisplay` 
* changed **tilt.py** by **level.py** as well as the names inside, and add a variable name Vlevel that may refer to Vtilt, VlevelZ, VlevelY or VlevelX
* in `RadarDisplay` changed scan_type to plot_type with names "radarPpi", "radarRhi" or "radarAirborne" to match "gridZ", "gridY" and "gridX" in  `GridDisplay` 
* add variable name Vcontainer that may refer to Vradar or Vgrid

### DIFF
--- a/artview/components/plot_radar.py
+++ b/artview/components/plot_radar.py
@@ -562,12 +562,12 @@ class RadarDisplay(Component):
         -----
             If Vradar.value is None, returns None
         '''
-        from .tools import interior
+        from .tools import interior_radar
         radar = self.Vradar.value
         if radar is None:
             return (np.array([]),)*7
 
-        xy, idx = interior(path, radar, self.Vtilt.value)
+        xy, idx = interior_radar(path, radar, self.Vtilt.value)
         aux = (xy[:, 0], xy[:, 1], radar.azimuth['data'][idx[:, 0]],
                radar.range['data'][idx[:, 1]] / 1000.,
                radar.fields[self.Vfield.value]['data'][idx[:, 0], idx[:, 1]],
@@ -625,6 +625,7 @@ class RadarDisplay(Component):
 
         limits = self.Vlims.value
         cmap = self.Vcmap.value
+
         if self.plot_type == "radarAirborne":
             self.display = pyart.graph.RadarDisplay_Airborne(self.Vradar.value)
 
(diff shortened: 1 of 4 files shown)


artview/components/level.py
```diff
@@ -0,0 +1,218 @@
+"""
+level.py
+
+Class instance used for modifying level via Display window.
+"""
```
nguy: My only real comment on the nomenclature changes is that we may want to add wording in some scripts that notes that levels = tilts for "radar" files.

gamaanderson: it is not much, but we have this [wording](https://github.com/gamaanderson/artview/blob/grid/artview/core/core.py#L56-L58)

gamaanderson: There is also wording [here](https://github.com/gamaanderson/artview/blob/grid/artview/components/level.py#L40)

artview/components/plot_radar.py
```diff
@@ -102,7 +102,7 @@ def __init__(self, Vradar, Vfield, Vtilt, Vlims=None, Vcmap=None,
         # Connect the components
         self.connectAllVariables()
 
-        self.scan_type = None
+        self.plot_type = None
 
         # Set plot title and colorbar units to defaults
         self.title = None
```
nguy: L149 still has a reference to ```Display```.

artview/components/level.py
nguy: Okay that second was what I was looking for.

nguy:
I had not tested the option such as ```python -m artview -F filename```. This returns an empty directory and fails to open on my machine.

nguy:
@gamaanderson If you resubmit, can you also replace:

standard.py
Line 8 ```def run(DirIn='./', filename=None, field=None):```

with ```def run(DirIn=os.getcwd(), filename=None, field=None):```

an import of os will also be required.

gamaanderson:
done! other update from my working branch came with it, but it is just grid related so I don't think it is a problem to merge it

nguy:
Awesome! Thanks for that.

Issue #47, opened 2015-07-02
Python 2.6 Troubles


pfhein:
Python 2.6 does not support the -m option the same way as Python 2.7.  `python -m artview` does not work but `python -m artview.__main__` works with an error.  That is only a documentation change for Python 2.6.

I think it works but the error I believe is a different problem.  The error it produces is the same as when I try to run it in the "installed folder".  This may not be due to Python 2.6 but the error below is the error that I am getting.

`denali:~$ python -m artview.__main__`
`/usr/local64/lib64/python2.6/site-packages/numpy/oldnumeric/__init__.py:11: ModuleDeprecationWarning: The oldnumeric module will be dropped in Numpy 1.9
  warnings.warn(_msg, ModuleDeprecationWarning)`
`./`
`Traceback (most recent call last):`
  `File "/usr/lib64/python2.6/runpy.py", line 122, in _run_module_as_main`
    `"__main__", fname, loader, pkg_name)`
  `File "/usr/lib64/python2.6/runpy.py", line 34, in _run_code`
   ` exec code in run_globals`
  `File "/home/software/usr/local64/lib64/python2.6/site-packages/artview/__main__.py", line 15, in <module>`
    `artview.run(DirIn, filename, field)`
  `File "/usr/local64/lib64/python2.6/site-packages/artview/scripts/standard.py", line 22, in run`
    `MainMenu = Menu(DirIn, filename, name="Menu") #initiate Vradar`
  `File "/usr/local64/lib64/python2.6/site-packages/artview/components/menu.py", line 47, in __init__`
   ` super(Menu, self).__init__(name=name, parent=parent)`
  `File "/usr/local64/lib64/python2.6/site-packages/artview/core/core.py", line 95, in __init__`
    `super(Component, self).__init__(parent=parent, flags=flags)`
`AttributeError: 'flags' is not a Qt property or a signal``

gamaanderson:
indeed this is not a python problem, but a PyQt one. It would be nice to know that version you are using. Anyway if you want to execute it is totally safe to remove flags from this [line](https://github.com/nguy/artview/blob/master/artview/core/core.py#L95) and if we can solve the problem I would have no problem in making this removal definitive

as for the -m option, my work around would be using -c, but that unfortunately don't have all the command line options

    python -c "import artview; artview.run()"

pfhein:
Here is the PyQt version number, 4.6.2-9.el6, so it is not the most 
[...]

Issue #53, opened 2015-07-13
User Check list


gamaanderson:
I would like to know how easily can an inexperience user go around artview. For that I've made the following list

- [x] can you understand and correctly use inline option -d, -f, -F and -s?
- [x] can you open a radar file?
- [ ] can you change files and sweeps with the arrow keys?
- [x] can you change sweep/field?
- [x] can you use zoom and pick-value?
- [x] can you change axes limits, title, units and colormap?
- [x] can you add/remove rings to the plot?
- [x] can you save a plot image?
- [ ] can you use ComponentControl to get the 2 displays showing diferent fields but always the same sweep? if you change the sweep from one display does the other change together?
- [ ] can you clear the Main menu from all widget, letting only the menubar?
- [ ] can you start a new 3th Display?
- [ ] can you find and use the pluging that executes pyart.correction.dealias_unwrap_phase? can you see the corrected velocity in the Display? can you get help on the options?
- [ ] can you start ROI (from the "Components" menu)?
- [ ] can you select a region with ROI?
- [ ] can you view, save and open the ROI table?

I don't know if we even have guinea pigs to try to go this check list, however I wanted to write it down!

nguy:
This is good to think about now. Do you view this list as a test of how intuitive ARTView is to use or if works on different systems? Or both?

As for intuitiveness: I'd be very interested to hear about this.
As for system stability:We have a few different systems that have been running the software (mostly for development purposes), but it's far from extensive.
I think a third thing to think about: How easy is it for someone to help development? Add a plugin.

Some of this will be important for the AMS Radar Conference. There is a short course focusing on the different open source radar tools and ARTView is going to be included. If you have any ideas I'd be interested to hear it.

gamaanderson:
I was concerned about intuitiveness, I'm pretty confident of stability as we use nothing more than basic PyQt.

I believe anyone knowing something of PyQt wouldn't have a problem with writing a plug-in and the tutorials should help to get started. If I have any ideas I will tell you

P.s I will be without of internet til Thursday

tjlang:
Since I haven't done a ton with ARTview for a while, I might be a good candidate for going thru this checklist, but I don't have a timeline as I am traveling the next couple weeks.

nguy:
@gamaanderson enjoy your time away.
@tjlang that would be great!

tjlang:
[...]

PR #63, opened 2015-08-03
Control

gamaanderson:
Before anyone else start addressing issue #53, here is somethings I already did:

* add tests for python 2.6, this should resolve issue #47 
* for python later than 2.7 change the menu layout to a tab layout, this is more intuitive
* some renaming in ComponentsControl, but not enough
* add script to run artview by just typing "artview"
* fix quick save file, although I don't understand what is the purpose as we already have savefile
* improve roi start (from the components menu) and add help button to roi

### DIFF
--- a/artview/components/plot_radar.py
+++ b/artview/components/plot_radar.py
@@ -778,9 +778,10 @@ class RadarDisplay(Component):
     ########################
     def _quick_savefile(self, PTYPE=IMAGE_EXT):
         '''Save the current display via PyArt interface.'''
-        PNAME = self.display.generate_filename(
+        path = self.display.generate_filename(
             self.Vfield.value, self.Vtilt.value, ext=IMAGE_EXT)
-        print "Creating " + PNAME
+        self.canvas.print_figure(path, dpi=DPI)
+        self.statusbar.showMessage('Saved to %s' % path)
 
     def _savefile(self, PTYPE=IMAGE_EXT):
         '''Save the current display using PyQt dialog interface.'''
@@ -788,7 +789,7 @@ class RadarDisplay(Component):
             self.Vfield.value, self.Vtilt.value, ext=IMAGE_EXT)
         file_choices = "PNG (*.png)|*.png"
         path = unicode(QtGui.QFileDialog.getSaveFileName(
-            self, 'Save file', '', file_choices))
+            self, 'Save file', PBNAME, file_choices))
         if path:
             self.canvas.print_figure(path, dpi=DPI)
             self.statusbar.showMessage('Saved to %s' % path)
--- a/artview/components/roi.py
+++ b/artview/components/roi.py
@@ -11,7 +11,7 @@ from matplotlib.path import Path
 from matplotlib.lines import Line2D
 import csv
 
-from ..core import Variable, Component, common, VariableChoose
+from ..core import Variable, Component, common, VariableChoose, componentsList
 
 
 class ROI(Component):
@@ -192,16 +192,30 @@ http://stackoverflow.com/questions/12608835/writing-a-qtablewidget-to-a-csv-or-x
         self.buttonOpenTable = QtGui.QPushButton('Open Tabular Data', self)
         self.buttonSaveTable = QtGui.QPushButton('Save Tabular Data', self)
         self.buttonResetROI = QtGui.QPushButton('Reset ROI', self)
+        self.buttonHelp = QtGui.QPushButton('Help', self)
         self.buttonViewTable.clicked.connect(self.viewTable)
         self.buttonOpenTable.clicked.connect(self.openTable)
         self.buttonSaveTable.clicked.connect(self.saveTable)
         self.buttonResetROI.clicked.connect(self.resetROI)
+        self.buttonHelp.clicked.connect(self.displayHelp)
 
         # Create functionality buttons
+
         self.rBox_layout.addWidget(self.buttonViewTable)
         self.rBox_layout.addWidget(self.buttonOpenTable)
         self.rBox_layout.addWidget(self.buttonSaveTable)
         self.rBox_layout.addWidget(self.buttonResetROI)
+        self.rBox_layout.addWidget(self.buttonHelp)
+
+    def displayHelp(self):
+        # XXX I invite anyone to improve this help information
+        text = "Draw a Path in the Display Window using the Mouse\n\n"
+        text += "Functions:\n"
+        text += "    Primary Mouse Button - add vertex\n"
+        text += "    Hold - freely draw path\n"
+        text += "    Secundary Button - close path\n"
+
+        common.ShowLongText(text)
 
     def viewTable(self):
         '''View a Table of ROI points'''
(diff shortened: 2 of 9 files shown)


artview/components/plot_radar.py
```diff
@@ -778,9 +778,10 @@ def _check_file_type(self):
     ########################
     def _quick_savefile(self, PTYPE=IMAGE_EXT):
         '''Save the current display via PyArt interface.'''
-        PNAME = self.display.generate_filename(
+        path = self.display.generate_filename(
             self.Vfield.value, self.Vtilt.value, ext=IMAGE_EXT)
-        print "Creating " + PNAME
+        self.canvas.print_figure(path, dpi=DPI)
```
nguy: I would prefer ```imagename``` to ```path``` because this generates a filename and not a filepath+name.

On that note, can you change ```path``` in L783 to ```os.path.join(os.getcwd(), imagename)```?

The reason this was in there was to quickly save an image to current working directory instead of selecting some directory and entering a filename.

artview/components/plot_radar.py
```diff
@@ -778,9 +778,10 @@ def _check_file_type(self):
...
+        self.statusbar.showMessage('Saved to %s' % path)
```
nguy: Glad you did this, I was making the same change, but I'll wait to address any other until we merge this.

artview/components/plot_radar.py
```diff
@@ -788,7 +789,7 @@ def _savefile(self, PTYPE=IMAGE_EXT):
             self.Vfield.value, self.Vtilt.value, ext=IMAGE_EXT)
         file_choices = "PNG (*.png)|*.png"
         path = unicode(QtGui.QFileDialog.getSaveFileName(
-            self, 'Save file', '', file_choices))
+            self, 'Save file', PBNAME, file_choices))
```
nguy: Where is PBNAME defined?

gamaanderson: PBNAME is defined in line 788, just above

artview/components/plot_radar.py
gamaanderson: if this is the only alteration you can do it while merging or in a subsequent commit

artview/components/roi.py
```diff
@@ -192,16 +192,30 @@ def CreateROIWidget(self):
...
         self.rBox_layout.addWidget(self.buttonResetROI)
+        self.rBox_layout.addWidget(self.buttonHelp)
+
+    def displayHelp(self):
+        # XXX I invite anyone to improve this help information
+        text = "Draw a Path in the Display Window using the Mouse\n\n"
+        text += "Functions:\n"
+        text += "    Primary Mouse Button - add vertex\n"
+        text += "    Hold - freely draw path\n"
+        text += "    Secundary Button - close path\n"
```
nguy: Let's try this method first and see how it works out.
When I draw a polygon, I don't need to hold down the button, do you?
I would suggest the following:

        text = "<b>Using the Region of Interest (ROI) Tool</b>\n\n"
        text += "Draw a path in the display window using the Mouse.\n"
        text += "Functions:\n"
+        text += "    Primary Mouse Button (e.g. left button)- add vertex\n"
+        text += "    Hold - freely draw path\n"
+        text += "    Secondary Button (e.g. right button)- close path\n"

nguy:
Have you tried creating a second status bar? Is this possible?

My thought is to stack 2 status bars at the bottom of each display. The upper status bar would be called something like ```infobar``` and could contain the filename, tilt, and field displayed.

artview/components/roi.py
gamaanderson: by holding one can draw a curve, not just a polygon

I'm totally open for wording, your is much nicer than mine. One Idea I had is that roi can still use some doc-strings in how it works, and we could use the doc-string itself, but that was just an idea, do what you think is better.

gamaanderson:
I don't think QMainWindow predict two status-bar, but one could force it somehow. Visually I don't like the idea, I already think the actual plot is too small.

nguy:
Plot being too small in terms of the Display size, area within the Display or the default limits?

artview/components/roi.py
nguy: To use the doc-string (in the code I assume) how would this work practically?

gamaanderson:
the visual area, it is probably less than 50% of the display window

nguy:
So I was just playing around with this. 

After I began messing around with various tools, I am no longer able to get the displays to change to new files via arrow keys or through Opening new file.

artview/components/roi.py
gamaanderson: `text = self.__doc__` would display the text that is [here](https://github.com/nguy/artview/blob/master/artview/components/roi.py#L18-L21)

gamaanderson:
the places to change would be [here](https://github.com/nguy/artview/blob/master/artview/components/plot_radar.py#L586-L587) and if plot_type changes [here](https://github.com/nguy/artview/blob/master/artview/components/plot_radar.py#L598-L599), however the problem is that re-size is proportional and font size is not. So if you put margins too small they will look good in a big display and bad in a small one.

nguy:
I wonder if we should leave this alone for now, I remember having issues with this when that first went into the program a ways back. I seem to recall also that the tick labels, etc. become an issue when messing with these values.

artview/components/roi.py
nguy: I could go either way on this one as well. I could certainly see the utility in just putting this string in the doc-string and pulling it in from there. Only issue might be if the doc-string continues to grow in the future...hmmm

gamaanderson: This doc-string is what will go to the documentation, so if it is in the documentation isn't it what the user need to know? the __init__ function has a string of its own.

nguy: Yes I see what you are saying. You are correct that ROI needs doc-string.

At this point the Help button is a nice addition though I think for a short explanation, just as you provided.

nguy:
@gamaanderson I'm going to go ahead an merge this and then make a couple modifications. I'll let you know when I put those through.

gamaanderson:
Ok, I still have a lot to be done in grid. Unfortunately I depend on pyart modifications.

nguy:
Okay, I think we can knock out a couple of small updates before that happens.

nguy:
@gamaanderson @tjlang 

I just committed some updates into a new push. Check out and let me know what you think. Nothing major here. Lots of housekeeping. But I did put an information text at the top:
screen shot 2015-08-03 at 5 10 46 pm

Changes include:
* Updated ROI help text.
* Update up/down arrow key help.
* Added an information field at the top of display.
* Updated some plot_grid functions.
* Minor change in quick_save.
* Extensive spelling/grammar updates.

gamaanderson:
Thank you very much for taking this @nguy.

about the questions:
1. yes, `set_limts` can be removed, as this is being done in `self._update_axes()`
2. the 'get' methods, are being use out-side display, in **roi.py** and **pick_value.py**. I would like to keep them as they are and add more just if needed

nguy:
Great, I can remove the set_limits on next go around. And keep the 'get' methods.

nguy:
Oh and I think we discussed that ```ROI class``` can be removed from ```tools.py```?

gamaanderson:
yes, it can

gamaanderson:
@nguy what a about a vertical colorbar, I will have to do it with the latitude/longitude slice to get a narrow display to work. I wounder if we could use it also in the radar display (screens now a days are normally longer in the horizontal axis)

nguy:
I've played with that before, but never liked it. The aspect ratio of the RHI and Airborne plots makes the vertical bar very squat.
That said it's my personal preference there and I expect this request some time down the road.

Hold off on changes for just a bit. I've got a PR I'll put out in just a bit that addresses a couple of points of @tjlang and @pfhein. I also made a couple of tweaks to LongMessage and ROI.

gamaanderson:
no problem, I must do a pyart PR before updating grid

Issue #65, opened 2015-08-04
Filename in Display


nguy:
@gamaanderson I want your opinion here. I need to bring the filename over from the menu (or Variable) class to the Display to assign in the new QLabel at top. Which do you think would be easiest?

I was thinking it may be as simple as passing a ```filename```  keyword to the ```plot_radar.py``` Display class?

gamaanderson:
no, passing this kind of information must be done with a variable, but I also wouldn’t what to add a Vfilename. You could in theory smuggle it in radar, by

    radar = pyart.io.read(filename)
    radar.filename = filename

and then test

    if hasattr(radar, "filename"):
       print radar.filename

gamaanderson:
But when one things get complicated, this is a signal that there is something wrong. In this case, artview is build around pyart, and pyart does not recognize the existence of files (Jonathan will say `Radar` is an abstract structure). Therefore ARTview in theory should follow the lead of py-ART and don't care about filename.

gamaanderson:
Just to be clear, I'm just say why of this uncomfortable limitation, not because of the Vvariable being a limitation, but this limitation is showing a concept conflict. I would also want `Radar` to be more closely related to disc files

nguy:
Consider it smuggled....for now.

nguy:
Everyone okay if I close this issue? It can be reopened if future change is desired.

gamaanderson:
ok, you may close it

PR #66, opened 2015-08-05
ROI and infolabel update.

nguy:
Added column headers to ROI saved CSV files.
Update open ROI saved files.
Enhanced error checking in ROI.
Added ROI stats output.
Fix LongMessage modeless to display.
Fixed a minor bug for Python 2.6
Added a tool reset button.
Updated infolabel.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,12 @@ To see the command line options:
 python artview -h
 ```
 
+If you are having trouble starting the program and you are running Python 2.6,
+try just using the following:
+```python
+artview
+```
+
 ARTview should be able to recognize and correctly handle PPI, RHI and airborne files.
 
 The default startup uses radar reflectivity and checks for a few common names.
--- a/artview/components/__init__.py
+++ b/artview/components/__init__.py
@@ -17,7 +17,7 @@ ARTview functions.
     LevelButtonWindow
     FieldButtonWindow
     LinkDisplay
-    ROI
+    SelectRegion
 """
 
 from .plot_radar import RadarDisplay
@@ -26,4 +26,4 @@ from .menu import Menu
 from .level import LevelButtonWindow
 from .field import FieldButtonWindow
 from .component_control import LinkDisplay
-from .roi import ROI
+from .select_region import SelectRegion
--- a/artview/components/__init__.py
+++ b/artview/components/__init__.py
@@ -16,7 +16,7 @@ ARTview functions.
     Menu
     LevelButtonWindow
     FieldButtonWindow
-    ComponentsControl
+    LinkDisplay
     ROI
 """
 
@@ -25,5 +25,5 @@ from .plot_grid import GridDisplay
 from .menu import Menu
 from .level import LevelButtonWindow
 from .field import FieldButtonWindow
-from .component_control import ComponentsControl
+from .component_control import LinkDisplay
 from .roi import ROI
--- a/artview/components/__init__.py
+++ b/artview/components/__init__.py
@@ -16,7 +16,7 @@ ARTview functions.
     Menu
     LevelButtonWindow
     FieldButtonWindow
-    LinkDisplay
+    LinkPlugins
     SelectRegion
 """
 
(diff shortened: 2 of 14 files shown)


nguy:
This takes care of a few comments from both @tjlang and @pfhein Issues #53 and #47 .
I also updated some ROI functionality, which led to a hopefully more robust modeless dialog window.

I also added a tooltip to self.canvas in ```plot_radar.py```, but I need suggestions on how to get the filename over the Display class. I balked at putting it in the new infolabel box at top after remembering how long file names can be. A subtle solution is a hover-over on the plot that will give the name. Of course this could cause problems with using the tools. Thoughts?

Maybe we could attach the tooltip to another feature?

gamaanderson:
About the filename, I am sure in my mind, the only `legal` way of passing it from Menu to Display is with a shared variable, however a Vfilename variable does not make sense, since I can not image of a use for component having Vradar unrelated to Vfilename. So if the official doors are closed, you smuggle it in. And radar is perfect for that.

nguy:
I spent some time floating through the code and looking online and yes it seems you are right given our setup. And I agree that a Vfilename seems unnecessary. As Tim pointed out, since the title is able to be changed (by design) it may not be the best way to tell what file you are reading at the moment. Plus, I've had instances of really not knowing from the auto-generated title. A quick reference would be nice here.

I may try the smuggle route with Vradar, attaching the filename as you previously suggest in ```Menu```. This can always be changed later if we deem it not worthy.

nguy:
Okay, updated the titles per conversation in #53 and also smuggled the filename into Display as discussed in #65 . The hover tool is now over the info label to see what file you are working on.

nguy:
I also made the change to LinkDisplay per #53 . Using the grep command it wasn't too bad, just need to be diligent in doing that.
Other changes I'm considering is ```RadarDisplay``` to ```AddDisplay``` and ```ROI``` to ```SelectRegion```. This might mean renaming ```roi.py``` to ```select_region.py``` also, but this is actually more clear. Thoughts?

pfhein:
Yes, SelectRegion is more descriptive than ROI.  I like the changes.

gamaanderson:
I don't like AddDisplay, Radar is important to separate from display, and add don't make sense as class name. If you want you can put add before all entries:
`"add " + comp.__name__` in  [here](https://github.com/nguy/artview/blob/master/artview/components/menu.py#L245)

nguy:
See #53 for my comment on AddDisplay, I agree and didn't change this.

Unfortunately I don't think "add" is appropriate as we're not really adding atteuation calculation they are being performed. And I like those names already.

I would go for ```LinkPlugins``` instead of ```LinkDisplay```.

gamaanderson:
would "start" be better than "add"?

nguy:
Okay, the check is added and will be pushed with next commit.
I also changed the name from ```LinkDisplay``` to ```LinkPlugins```. Does that work?

gamaanderson:
yes, it does

nguy:
Okay, the points.py has been added to core directory. I'm going to go ahead an merge.

gamaanderson:
Ok, go on and merge. By the way are you planning any more alterations?

nguy:
I will hold off for a bit. Are you going to implement something?

gamaanderson:
May be just some minor stuff to further round the edges.

nguy:
Sounds good. I have a couple of thoughts of things to do. One more addition
to SelectRegion and make the masking plugin.

gamaanderson:
Do you want to do that before or after I convert SelectRegion to use Points?

nguy:
Let me see if I can knock this out in the next day or two. If not I'll post here again.

Using the points won't change the functionality of SelectRegion, correct?

gamaanderson:
correct, the functionality will not change, just how the data is stored.

nguy:
I'm going to make just a couple quick changes and then I'll do another small PR.

I *may* have found a slightly more elegant test solution for the ```_check_default_field``` in ```plot_radar.py```. We'll see what others think. It came out of a conversation with @pfhein about how another software works.

PR #67, opened 2015-08-07
Simple plot class, histogram for SelectRegion.

nguy:
Added a simplified display, PlotDisplay, for basic plotting.
Added histogram plotting to SelectRegion.
Minor code cleanup.

### DIFF
--- a/artview/components/__init__.py
+++ b/artview/components/__init__.py
@@ -18,6 +18,7 @@ ARTview functions.
     FieldButtonWindow
     LinkPlugins
     SelectRegion
+    PlotDisplay
 """
 
 from .plot_radar import RadarDisplay
@@ -27,3 +28,4 @@ from .level import LevelButtonWindow
 from .field import FieldButtonWindow
 from .component_control import LinkPlugins
 from .select_region import SelectRegion
+from .plot_simple import PlotDisplay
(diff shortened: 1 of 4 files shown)


nguy:
@gamaanderson pending you and others comments and merger, I'll pass the torch to you this weekend.

I went with a quick and simplified class here for basic plotting. It isn't meant to be quite as nice as the Radar/GridDisplay, but to help in analysis.

gamaanderson:
@nguy I believe you forgot to add the new file. Also, have you add it to `artview/components/__init__.py`?

nguy:
@gamaanderson It's been a long week!  It's there now.
Thanks!

gamaanderson:
ok, it is clear that this was done in a simplify matter (no problem there) and there is some room for improvement, we main even pass a part of it together with **points.py** as a suggestion to pyart. 

However since it is working, I have no problem with merging as it is. That is because it is a new component/plugin, and plugins writers have total liberty to write the code as they want to, as long as they respect the over all structure. That is why I would merge it now and make questions later, and since we are looking forward for a release "later" really means "later".

nguy:
Yep, this was a simple program to get a foundation laid for future development. I've actually been thinking that converting this to a more advanced graphics class like RadarDisplay could yield some neat stuff in the future. 
But, I'll go ahead and merge and then we can create some issues to further this along.

PR #71, opened 2015-08-10
Points

gamaanderson:
Convert SelectRegion to use of Points.


artview/components/plot_radar.py
```diff
@@ -586,10 +587,10 @@ def getPathInteriorValues(self, path):
...
         Returns
         -------
-        x, y, azi, range, value, ray_idx, range_inx: ndarray
-            Truplet of 1arrays containing x,y coordinate, azimuth,
-            range, current field value, ray index and range index
-            for all bin of the current radar and tilt inside path.
+        points: Points
+            Points object containing all bins of the current radar
+            and tilt inside path. Axes : 'x_disp', 'y_disp', 'ray_index',
+            'range_index', 'azimuth', 'range'. Fields: just current field
```
nguy: I like the fact that you included the field in here. I had trouble figuring out how to do this.

artview/components/select_region.py
```diff
@@ -247,44 +247,17 @@ def viewTable(self):
...
-#                for i in range(self.ind.size):
-#                warn = common.ShowWarning(msg)
-#                    print "%8.2f %8.2f %8.2f %8.3f %8.2f %8d %8d" %\
+        points = self.Vpoints.value
+        if points is not None:
+            fsuggest = 'SelectRegion_' + self.getField() + '_' + \
+                str(points.axes['x_disp']['data'][:].mean()) + '_' + \
+                str(points.axes['y_disp']['data'][:].mean())+'.csv'
+            path = QtGui.QFileDialog.getSaveFileName(
+                    self, 'Save CSV Table File', fsuggest, 'CSV(*.csv)')
```
nguy: Does fsuggest work for you? It does nothing on my system, which is why I abandoned it a while back.

artview/components/select_region.py
```diff
@@ -247,44 +247,17 @@ def viewTable(self):
...
+            if not path.isEmpty():
+                write_points_csv(path, points)
+        else:
+            common.ShowWarning("Points is None, the is nothigh to save!")
```
nguy: Should read: "Points is None, no data to save!"

artview/core/common.py
```diff
@@ -155,42 +155,49 @@ def startDisplay(self):
...
-# Table methods #
+#    Table methods     #
 ########################
 
 
 class CreateTable(QtGui.QTableWidget):
     """Creates a custom table widget."""
-    def __init__(self, column_names, name="Table",
+    def __init__(self, points, name="Table",
                  textcolor="black", bgcolor="gray", parent=None, *args):
```
nguy: Do you think the ```points``` structure is more flexible for displaying other data as well?

artview/components/plot_radar.py
```diff
@@ -598,14 +599,46 @@ def getPathInteriorValues(self, path):
...
+                   'long_name': 'index in ray dimension'}
+        rng_idx = {'data': idx[:, 1],
+                   'long_name': 'index in range dimension'}
+
+        axes = {'x_disp':xaxis,
+                'y_disp':yaxis,
+                'ray_index':ray_idx,
+                'range_index':rng_idx,
+                'azimuth':azi,
+                'range':rng}
```
nguy: I wonder if we should include the units in a couple of these because they are different than the display which makes the adjustment automatically (e.g. meters to kilometers):
```
        axes = {'x_disp_m':xaxis,
                'y_disp_m':yaxis,
                'ray_index':ray_idx,
                'range_index':rng_idx,
                'azimuth':azi,
                'range_m':rng}```

File: artview/core/common.py [hunk 6]
```diff
@@ -155,42 +155,49 @@ def startDisplay(self):
...
+        self.setHorizontalHeaderLabels(colnames)
 
         for i in xrange(nrows):
             # Set each cell to be a QTableWidgetItem from _process_row method
-            for j in xrange(ncols):
-                item = QtGui.QTableWidgetItem(str(data[i, j]).format("%8.3f"))
+            for j, name in enumerate(colnames):
+                if name in self.points.axes:
+                    item = QtGui.QTableWidgetItem(str(
+                        self.points.axes[name]['data'][i]).format("%8.3f"))
```
nguy: I'm not sure this format code is being applied to x_axis and y_axis, when I put up table, there are at least 6 decimal digits. The others display fine, any idea why this is?

artview/components/select_region.py
gamaanderson: Yes it works, but that kind of thing is really system dependent. For instance -d option don't work with relative path in KDE, but it does in GNOME

artview/components/select_region.py
gamaanderson: ok, I will change that.

artview/core/common.py
gamaanderson: It depends in the uses that we will give to it. If you think it will be more usefully the way it was we may change it back, but I will then have to convert Points to data array.

artview/components/plot_radar.py
gamaanderson: I'm using meter because that is the default of Grid, the keys also come from Grid. Any way the right way (CF) of informing the units in in the 'units' attribute, not the name.

artview/components/select_region.py
nguy: Very interesting. Tough to keep track of all of the issues with different systems.

artview/core/common.py
nguy: I like the idea behind it and it really cleans up a few parts of the code that bugged me (for example reading the 4th dimension of the data array to calculate the mean). I feel this is the way to go, but there may be potential tweaks down the road.

artview/components/plot_radar.py
nguy: But the issue here is that the table columns use the key names. Just wondering if this will cause any confusion to those less familiar with radar? I'm flexible on this as the previous names also did not make this clear.

I hesitate producing CF files for the Points (especially SelectRegion), but maybe this is no worse than CSV files.

gamaanderson: I have listed which axes  are allowed in the Points, so we may change the table labels based in a dict.

artview/core/common.py
gamaanderson: seeing it now I don't think this should work, this is equivalent to `'1.123'.format("%.1f") = '1.1'`, that does not happen. I will re-write in a may that should work (using `'%.3f' % value`.)

PR #72, opened 2015-08-10
Bug fix in components/plot_radar.py

tjlang:
In components/plot_radar.py, there is an error introduced when working
with NEXRAD files that have been converted away from Level 2 by
RadxConvert and have had their split-cut sweeps merged. When looping
over self.rTilts in RadarDisplay, if you do not loop over the array
indices themselves, you will get an IndexError and artview will crash
because sweep numbers and index numbers do not match. I fixed this with
enumerate().

During the error tracking process, I also applied pep8 to following:
__main__.py (did not address pep8 complaint about location of artview
import)
scripts/standard.py
components/plot_radar.py

### DIFF
--- a/artview/__main__.py
+++ b/artview/__main__.py
@@ -2,12 +2,12 @@
 import sys
 import os
 
-
 path = os.path.dirname(sys.modules[__name__].__file__)
 path = os.path.join(path, '..')
 sys.path.insert(0, path)
 
 import artview
+
 
 def main(argv):
     script, DirIn, filename, field = artview.parser.parse(argv)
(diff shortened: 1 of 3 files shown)


gamaanderson:
thank you for catching this @nguy. We really should not use a variable (sweep number) for index.

nguy:
Nice catch Tim. 
I noticed that the Custom Tool is still in the Display Tools list. I'm thinking that we could have a plugin called CreateCustomPlugin that brings up a text window with  information on creating a plugin and the website with the directions. Any objections?

gamaanderson:
The idea of this  CreateCustomPlugin its kind of strange to me, its not like one could create a plugin in just as few lines. Place that in the Display is also kind of not the right place. Any way I wouldn’t say I have objections, I'm not voting it down in any way, I just find it really strange.

nguy:
If we don't think it's the right place that's fine. Maybe it's better to keep this to the documentation and not try to confuse things.

PR #78, opened 2015-08-12
ENH: New Py-Art feature

gamaanderson:
As we are now using pyart 1.5 there are some new features that we can use: default colormap in `pyart.config` and modular `generate_title`.
This is an enhancement, merge when you find most appropriate.


nguy:
Great @gamaanderson ! These are definitely worth pushing into the latest!

PR #82, opened 2015-08-22
Drag drop

gamaanderson:
I would like to NOT merge this just jet, but let as parallel development til we get it right

It is not looking really good, but the drag-drop link/unlink is already working and the `DropDown_DealiasRegionBased` exemplify how this can be used to replace the `VariableChoose`. I would like some comment on how to improve the user side of this.

Notice that the drag-drop operation are done in core, not in components. One may try to change my mind, but I believe to be important to make this work out of the box and not demand from the component's programmer to build it.

### DIFF
--- a/artview/components/plot_radar.py
+++ b/artview/components/plot_radar.py
@@ -9,6 +9,8 @@ import os
 import pyart
 
 from PyQt4 import QtGui, QtCore
+import matplotlib
+
 from matplotlib.backends.backend_qt4agg import FigureCanvasQTAgg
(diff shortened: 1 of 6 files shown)


nguy:
I'll take a look at this as soon as I can.

gamaanderson:
small update: I tested this with my roommate and it was absolutely use-less, the current process was much better!

gamaanderson:
I'm closing this PR. Lets say it failed the proof of concept. I have no further intention of developing this.

nguy:
You got it. If it rears it's head later, that's fine.

PR #84, opened 2015-08-27
Terminal Access

gamaanderson:
Add a plugin to give the user direct access to artview, so he can change the data directly in python. It is a powerful tool for some one that knows what he is doing, but it can be dangerous as well if not used carefully. There may be some ways of making it safer, but it involves much more code for emulating a terminal, however the idea itself of giving a terminal to someone is risk.

### DIFF
--- a/artview/plugins/console.py
+++ b/artview/plugins/console.py
@@ -1,0 +1,87 @@
+"""
+console.py
+"""
+
+# Load the needed packages
+from PyQt4 import QtGui, QtCore
+
+import code
+import pyart
+
+import sys
+import os
+
+path = os.path.dirname(sys.modules[__name__].__file__)
+path = os.path.join(path, '...')
+sys.path.insert(0, path)
+
+import artview
+
+from .. import core
+common = core.common
+
+
+class AccessTerminal(core.Component):
+    '''
+    Open an interactive python console so the direct manipulation
+    '''
+
+    @classmethod
+    def guiStart(self, parent=None):
+        '''Graphical interface for starting this class.'''
+        kwargs, independent = \
+            common._SimplePluginStart("AccessTerminal").startDisplay()
+        kwargs['parent'] = parent
+        return self(**kwargs), independent
+
+    def __init__(self, name="AccessTerminal", parent=None):
+        '''Initialize the class to create the interface.
+
+        Parameters
+        ----------
+        [Optional]
+        name : string
+            Field Radiobutton window name.
+        parent : PyQt instance
+            Parent instance to associate to this class.
+            If None, then Qt owns, otherwise associated with parent PyQt
+            instance.
+        '''
+        super(AccessTerminal, self).__init__(name=name, parent=parent)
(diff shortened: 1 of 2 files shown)


nguy:
This is part of the reason I included the ```Show Script``` in the upcoming gatefilter (still working on this on the pyart side). This way we could produce the artview calls that could then be used in a terminal window. It definitely has it's risks, but unless the user save the newly produced data I am inclined to say it's benign.

That being said, I have no problem exploring this further.

nguy:
Is this something you'd like to merge or wait?

gamaanderson:
This is more than given a script, it give total control over artview at execution time, one easily open a new display, a new radar or modify the existing ones and seeing the result direct on the running artview.

I think this can be merged.

artview/plugins/console.py
```diff
@@ -0,0 +1,87 @@
...
+        print artview
+        banner = ("HELLO: this is an iteractive python console so you can\n" +
+                  "acess ARTview from the incide and manipulate the data\n" +
+                  "directly. You have acess to two variables:\n" +
+                  "    'components': List of all running components\n" +
+                  "    'pyart': Python-ARM Radar Toolkit.\n" +
+                  "    'artvew': ARM Radar Toolkit Viewer\n" +
+                  "To leave and go back to grafical ARTview press ctrl+D\n" +
+                  "in unix and OXS or ctrl+Z in Windows.")
+        thread = threading.Thread(
```
nguy: Can we change banner to the following?

```
        banner = ("HELLO: this is an iteractive python console so you can\n" +
                  "access ARTview functions while running the GUI and manipulate the data directly.\n\n" 
                  " You have acess to two variables:\n" +
                  "    'components': A list of all running components\n" +
                  "    'pyart': Python-ARM Radar Toolkit.\n" +
                  "    'artview': ARM Radar Toolkit Viewer\n\n" +
                  "To leave and go back to graphical ARTview press ctrl+D\n" +
                  "in Unix/OXS or ctrl+Z in Windows.")

File: artview/plugins/console.py [hunk 2]
```diff
@@ -0,0 +1,87 @@
...
+                    'readfunc': None,
+                    'local': {'components': core.componentsList, 'pyart': pyart,
+                              'artview': artview}}
+            )
+        thread.start()
+        thread.join()
+
+
+
+_plugins = [AccessTerminal]
```
nguy: Can we also add an exit message to print to terminal:
```print("Ending interactive session, adios!"

Or something equally cheeky. :)

artview/plugins/console.py
gamaanderson: yes, of course, just correct for me "two" to "three variables"

artview/plugins/console.py
gamaanderson: as you wish.

artview/plugins/console.py
nguy: Done and merging.

Issue #80, opened 2015-08-18
crash after opening file


vanandel:
With conda 3.16.0 on OX 10.10.5, installed latest 
artview : de80c70b6f15d4915153f286a79bd92450e94b4a

and pyart: a07b186fd7b9603d5e718daef1ad088bc2510567

I opened a CFRadial file and got the following crash:
Opening file /scr/eldora2/rsfdata/front/cfradial/spol/qpe/20150404/cfrad.20150404_224032.000_to_20150404_224032.000_SPOL_v664_s00_el85.00_SUR.nc
Traceback (most recent call last):
  File "/usr/local/anaconda/bin/artview", line 5, in <module>
    artview.__main__.main(sys.argv)
  File "/usr/local/anaconda/lib/python2.7/site-packages/artview/__main__.py", line 18, in main
    artview.run(DirIn, filename, field)
  File "/usr/local/anaconda/lib/python2.7/site-packages/artview/scripts/standard.py", line 50, in run
    parent=MainMenu)
  File "/usr/local/anaconda/lib/python2.7/site-packages/artview/components/plot_radar.py", line 137, in __init__
    self.LaunchGUI()
  File "/usr/local/anaconda/lib/python2.7/site-packages/artview/components/plot_radar.py", line 166, in LaunchGUI
    self._set_figure_canvas()
  File "/usr/local/anaconda/lib/python2.7/site-packages/artview/components/plot_radar.py", line 672, in _set_figure_canvas
    self.layout.addWidget(self.canvas, 1, 0, 7, 6)
TypeError: arguments did not match any overloaded call:
  QGridLayout.addWidget(QWidget): argument 1 has unexpected type 'FigureCanvasQTAgg'
  QGridLayout.addWidget(QWidget, int, int, Qt.Alignment alignment=0): argument 1 has unexpected type 'FigureCanvasQTAgg'
  QGridLayout.addWidget(QWidget, int, int, int, int, Qt.Alignment alignment=0): argument 1 has unexpected type 'FigureCanvasQTAgg'

gamaanderson:
@vanandel Thank you for reporting this bug.

Apparently (py)Qt fail to recognize matplotlib canvas as a Widget, I'm not sure why is that, since `matplotlib.backends.backend_qt4agg.FigureCanvas` is a subclass of `QtGui.QWidget` and for me `isinstance(QtGui.QWidget, self.canvas)` returns `True` as expected.

The following information could help me in debugging this:
python version
Qt version
PyQt version
[...]

PR #92, opened 2015-09-13
Qt Backend

gamaanderson:
I implemented the solution of issue #80, but I couldn't hold my self and did pep8, doc update and other modification here and there. To avoid the huge diff see the individual commits.

Don't merge still, there is still some cleaning to be done, and by the way ,should we let the "Toggle GateFilter" as a radio menu or try the checkbox?


nguy:
We can try the checkbox if you like, I have no strong feeling currently on this.

Thanks for the updates @gamaanderson - I'm curious are the less explicit links pep8 or something you did to clean up?

gamaanderson:
You will need to be more specific in your question, but I most say the commits are right separated, so if it is not in commit 5be6ffc it was probably my decision.

gamaanderson:
please see the new commits, gatefilter toggle is much cleaner now.

Issue #97, opened 2015-09-24
Manual Unfold


gamaanderson:
This I idea just came to me: add a plugin for manual velocity unfolding, that is, the user select a region (with the `SelectRegion` plug-in or with some cluster identification tool) that he knows is folded and we unfold it! Using a `SelectRegion` it is pretty easy, just need to link Vradar and Vpoints and perform the operation.

@jjhelmus, this is probably a pure ARTview functionality, but I just wanted to let you know, may be you have some idea that we can use/contribute to pyart.

jjhelmus:
Nick and I discussed this at the AMS conference last week and it was a commonly requested feature of ARTView.  What would be nice in such a plugin would be to have a button/option to first perform unfolding using one of Py-ART's dealiasing routines and then allow hand unfolding of the results.  The idea here being that the automatic method can get you a long ways but the resulting volume may still need some minor modifications.  Exposing the regions used in the region based dealiasing method to the user to that they could easily select and unfold a region where velocities are similar would also be helpful.

gamaanderson:
There are already some plug-ins implementing pyart unfolding, so this already partially done. I could really fast implement a simple manual version, but I think using it will probably be "uncomfortable". I would really like to user could just click in a region and we identify the borders. Also if we could separate the already implemented pyart methods in two parts: identification and unfolding, we could try to show/modify the identification results (GateFilter ?) before applying. Would this separation be doable?

jjhelmus:
Definitely doable, and in fact already partially implemented.  The `dealias_region_based` function unfolds a volume sweep by sweep.  Identification of regions of similar velocities is accomplished using the [_find_regions](http://arm-doe.github.io/pyart-docs-travis/dev_reference/generated/pyart.correct.region_dealias._find_regions.html#pyart.correct.region_dealias._find_regions) function.  This function requires numpy arrays which would need to be extracted from the Vradar but should be doable.  Injecting this back into the unfolding algorithm would take a bit more work.

nguy:
Doh, I forgot to mention this one in the email! Thanks for bringing it up @gamaanderson 

I wonder if adding a button for the manual unfolding to the pyart plugins would be the way to go. That way the regions that do not get completely unfolded can be dealt with by popping up a SelectRegion instance.

gamaanderson:
[...]

PR #100, opened 2015-10-03
add manual unfold plugin and script

gamaanderson:
This manual unfold plugin is waiting some feed back for improvement (issue #97). It is probably some what difficult to work with.


nguy:
I'll try to get to this later this week. I'm loaded up with obligations.

nguy:
Oh, I'm going to go ahead and merge the SelectRegion this week as well.

nguy:
I think there needs to be a help button with explanation on how to use this.

nguy:
After linking the radar instance from ```Display1``` and ```ManualUnfold``` I set the Nyquist velocity and clicked both Pos/Neg buttons. An error of "Radar is None" appears.

nguy:
Another interesting thing is that after selecting the ```ManualUnfold``` the Display1 Tilt and Field drop down menus are no longer populated.

nguy:
Ah, okay so I had to re-link the Menu and Display1 instances.

gamaanderson:
You probably linked `Display`->`ManualUnfold`, instead of `ManualUnfold`->`Display`.

nguy:
Doesn't seem to make a difference how I link. No luck with any when linked.

gamaanderson:
Isn't the error "Points is None"? you also have to link to the SelectRegion to get the points

artview/plugins/manual_unfold.py
```diff
@@ -0,0 +1,176 @@
...
+        radar.add_field_like(vel, corrVel, data, replace_existing=True)
+        self.Vradar.update(strong_update)
+
+    def newRadar(self, variable, value, strong):
+        '''respond to change in radar.'''
+        if self.Vradar.value is None:
+            return
+
+        nyquist_vel = self.Vradar.value.get_nyquist_vel(0, True)
+        self.nyquistVelocit.setValue(nyquist_vel)
```
nguy: This should be ```self.nyquistVelocity.setValue(nyquist_vel)```, the "y" is missing.

gamaanderson: you are right, but this just break the automatic recognition of nyquist_vel, the plugin should still be working.

nguy:
Okay my latest testing last night after selecting ```ManualUnfold``` and ```SelectRegion``` and linking was still unsuccessful with the error that Vpoints was None. I'm having trouble tracking down why this is happening.

Also unable to run the script on a directory with NEXRAD data. I get the following:
```python
guy:20130520 $ artview -s manual_unfold
Opening file /Users/guy/data/nexrad/KTLX/nex2/20130520/KTLX20130520_235401_V06
2015-10-23 08:52:59.209 python[59532:30916062] modalSession has been exited prematurely - check for a reentrant call to endModalSession:
Warning Discarded!
  File "/Users/guy/anaconda/bin/artview", line 5, in <module>
    artview.__main__.main(sys.argv)
  File "/Users/guy/anaconda/lib/python2.7/site-packages/artview/__main__.py", line 16, in main
    artview.scripts.scripts[script](DirIn, filename, field)
  File "/Users/guy/anaconda/lib/python2.7/site-packages/artview/scripts/manual_unfold.py", line 41, in run
    roi = SelectRegion(plot1, name="SelectRegion", parent=menu)
  File "/Users/guy/anaconda/lib/python2.7/site-packages/artview/components/select_region.py", line 107, in __init__
    exclude_based=True)
  File "/Users/guy/anaconda/lib/python2.7/site-packages/pyart/filters/gatefilter.py", line 161, in __init__
    shape = (radar.nrays, radar.ngates)
AttributeError: 'NoneType' object has no attribute 'nrays'
```

I also get the error "Could not open file, Invalid mode!"

gamaanderson:
I can see now that your local branch is merged, this PR is some what old and apparently incompatible with some of the updates in master. I will make the merge in my side and correct the problems, I don't know if this will create conflicts in your side.

nguy:
Okay, sorry about that, it didn't appear conflicts would be a problem.

nguy:
@gamaanderson I am looking into the new submissions and notice that when selecting Advanced Tools, the Window Initialization always has 'PhaseProcLp' in it instead of the Plugin names.

nguy:
The SelectRegion was unable to start from the Display Toolbox.

So did the SelectRegion buttons move elsewhere? You mention a ```PointsDisplay``` that may exist?

artview/components/select_region.py
```diff
@@ -40,35 +41,38 @@ class SelectRegion(Component):
...
 
     @classmethod
     def guiStart(self, parent=None):
-        args, independent = _SelectRegionStart().startDisplay()
-        return self(**args), independent
+        #args, independent = _SelectRegionStart().startDisplay()
+        kwargs, independent = \
+            common._SimplePluginStart("AccessTerminal").startDisplay()
+        kwargs['parent'] = parent
+        return self(**kwargs), independent
```
nguy: I don't follow the change here. Why the "AccessTerminal" start implementation?

artview/components/select_region.py
```diff
@@ -40,35 +41,38 @@ class SelectRegion(Component):
...
 
-    def __init__(self, display, name="SelectRegion", parent=None):
+
+    def __init__(self, VplotAxes=None, VpathInteriorFunc=None, Vfield=None,
+                 name="SelectRegion", parent=None):
         '''
```
nguy: By adding the ```pathInteriorFunc```, etc here, you are now able to not have SelectRegion dependent upon an individual Display, correct?

artview/components/select_region.py
```diff
@@ -13,7 +14,7 @@
 
 from ..core import (Variable, Component, common, VariableChoose,
                     componentsList, QtGui, QtCore)
-from ..core.points import Points, write_points_csv, read_points_csv
+from ..core.points import Points
```
nguy: Is there still a way to save the points to a document? Or are you thinking that having a list of polygons will allow the same functionality?

artview/components/select_region.py
```diff
@@ -145,7 +141,7 @@ def motion_notify_callback(self, event):
             elif event.button == 1:  # Free Hand Drawing
                 line = Line2D([self.previous_point[0], x],
                               [self.previous_point[1], y])
-                self.poly.append(ax.add_line(line))
+                self.polys[-1].append(ax.add_line(line))
                 self.previous_point = [x, y]
```
nguy: A list of lists to store the polygons? Clever solution.

artview/components/select_region.py
```diff
@@ -155,16 +151,19 @@ def button_press_callback(self, event):
         Grab the data when line is drawn.
         If shape is closed, then compile points within.
         '''
+        # test for double click support (matplotlib > 1.2)
+        db_support = 'dblclick' in dir(event)
         if event.inaxes:
```
nguy: Is this the same across different OS?

artview/components/select_region.py
```diff
@@ -173,39 +172,52 @@ def button_press_callback(self, event):
                                        marker='o')
                     self.previous_point = [x, y]
```
nguy: We may want to decrease the marker size? I find them kind of large to work with.
And one issue I would love to make happen is to not have to "close" the polygon fully, but have an automatic closure if an action was chosen.

artview/components/select_region.py
```diff
@@ -218,48 +230,40 @@ def CreateSelectRegionWidget(self):
         self.setCentralWidget(self.SelectRegionbox)
 
         # Add buttons for functionality
-        self.buttonViewTable = QtGui.QPushButton('View Tabular Data', self)
```
nguy: So where will these functions be? I think they are quite useful.

artview/components/select_region.py
```diff
@@ -218,48 +230,40 @@ def CreateSelectRegionWidget(self):
...
         self.buttonHelp.setToolTip("About using SelectRegion")
-        self.buttonViewTable.clicked.connect(self.viewTable)
-        self.buttonOpenTable.clicked.connect(self.openTable)
-        self.buttonSaveTable.clicked.connect(self.saveTable)
-        self.buttonStats.clicked.connect(self.displayStats)
-        self.buttonHist.clicked.connect(self.showHist)
-        self.buttonApplyMask.clicked.connect(self.applyMask)
-        self.buttonRestoreMask.clicked.connect(self.restoreMask)
+        self.buttonBack = QtGui.QPushButton('Back', self)
+        self.buttonBack.setToolTip("Remove last Polygon")
```
nguy: Should this be labeled something other than 'Back'? This removes the current or last polygon, but back would suggest being able to scroll through the polygon list.

artview/core/core.py
```diff
@@ -165,6 +173,17 @@ def __repr__(self):
...
+    '''Sugest a unambiguous name for component.'''
+    compName = comp.__name__
+    name = compName
+    i = 0
+    names = [c.name for c in componentsList]
+    while name in names and i < 1000:
+        name = compName + '%i' % i
+        i = i + 1
+    return name
+
```
nguy: Where is this used?

gamaanderson:
I totally forgot the display also has a (code inter-dependent) reference to `SelectRegion`. For now that is just broken.

`PointsDisplay` does not exist, it will exist. However the start model for that, is the plot_simple you made.

artview/components/select_region.py
gamaanderson: this is a bug!

artview/components/select_region.py
gamaanderson: no, that was already the case (or not depending in your interpretation). I had to change it here to get it in the "Variable" standard, so the new "modes" work (I have said it, all variables must be optional ones)

artview/components/select_region.py
gamaanderson: this will still be possible, but in `PointsDisplay`, no longer here.

artview/components/select_region.py
gamaanderson: yes, I had to do it.

artview/components/select_region.py
gamaanderson: this is dependent of matplotlib version, not OS. If matplotlib is correctly interpreting the OS, is an other question, but the right click close is still working.

artview/components/select_region.py
gamaanderson: See how your image editor work, those softwares are the specialists in this functionality.

artview/components/select_region.py
gamaanderson: `PointsDisplay`

artview/components/select_region.py
gamaanderson: It depends how much time one spend creating a polygon, if it is to quick that is not needed, otherwise we could implement. As always I'm open for name suggestions.

artview/core/core.py
gamaanderson: `Menu.change_mode`

artview/components/select_region.py
nguy: Got it.

artview/components/select_region.py
nguy: Will it be possible to start a PointsDisplay from SelectRegion? I think we should consider this.

I hate to lose the functionality. Will this be added before we finalize a first round of manual_unfold?

artview/components/select_region.py
nguy: Great, that explains the matplotlib note.

artview/components/select_region.py
nguy: See above about PointsDisplay.

artview/components/select_region.py
nguy: Let's change to ```RemovePoly```,  which I think is descriptiveor alternatively ```Undo```, and restore to the previous step.
The Solo editor, which we are working slowly to replace stores around 5 previous polygons. We should think about something like this (I don't have a specific number in mind). @deeplycloudy asked about this functionality.

artview/core/core.py
nguy: Okay. Nitpick but can you change `sugest` to `suggest` for these instances?

nguy:
Okay, this last answers my previous questions. I think we should probably have ```PointsDisplay``` going when we merge this so we don't lose capabilities.

artview/components/select_region.py
gamaanderson: I wouldn't like a button in SelectRegion, I was thinking more about a "extract points" mode that open and link SelectRegion and PointsDisplay.

If we want to merge it before PointsDisplay is ready, we could allow two parallel SelectRegion, I have the old file as back up.

artview/components/select_region.py
gamaanderson: I would write `"Remove Polygon"` it is more verbose.

artview/core/core.py
gamaanderson: yes, no problem.

artview/components/select_region.py
nguy: I'm open to that, maybe under "Points Data" or something (whatever is fine). I like the idea of linking to both. 

Let's have parallel SelectRegion (could even rename SelectPoints or something if we need another name) so that we have both ongoing and then narrow down the scope through future commits. What do you think?

gamaanderson: If you want a fast merge we can do it.

I will just take some time till I get back to the code for correcting some of the points here. I will open a PR when it is ready.

nguy: I'm in no big rush to merge. I think that we can take our time and correct these points first.

gamaanderson:
just push corrections from bugs found here. The new version of SelectRegion can be found by SelectRegion_dev, the old one is back to its place. There reason i choose this name, is that this is intended to replace SelectRegion in the future and not to exist in parallel.
In the menu you can found the new version in the manual_unfold mode.

nguy:
I've been playing with this a little. All the issues seem to be fixed that went awry. I think your 'dev' naming convention is a fine idea. And as this matures it can then be phased out without any issues.

I am apt to merge this PR for now and then we can do incremental updates as needed. Thoughts?

gamaanderson:
did you saw the latest version I uploaded today?

nguy:
Oh, no I only saw the one from about 2 weeks ago. Let me take a look.

nguy:
Hmmm, so when i select a region of folded velocity and try the manual unfold I don't see any effect.
Also, I'm not clear about the linking. I tried linking points and radar to ```SelectRegion``` but the radar was None and threw and error. Then linked the radar variable from ```Display1``` to ```ManualUnfold``` and *something* happened, but no obvious effect.

Is there a help or example to explain useage?

gamaanderson:
The two last commits don't change the manual unfold, it is in the same stand as for 2 weeks. New are the new extract_points and manual_filtering modes.

Manual linking also shouldn’t be needed any more, the modes menu should perform all needed linking.

I have no example, it should be pretty straight forward. Sorry for the little help.

nguy:
Okay, by using the explicit modes everything links seemingly. But when I apply an unfold it appears that no action is taken. Well that's not true, there is a delay and obviously something is happening, but no values change. This is how it works, correct?

gamaanderson:
try to look for a new corrected field.

nguy:
Loaded it and no  change.

gamaanderson:
so you found a new field in your radar, but it is equal to the original velocity?

nguy:
Oh wait a minute! The change was so small I missed it the first time. There are changes be applied!

nguy:
And by the way, ARTView kicks ass! I just linked the limits on the 2 displays to easily compare a region. :)

gamaanderson:
:)

nguy:
Here's a potential bug though. I'm trying to modify the Data Limits and it's bombing. Goes from 0-1 and acts like there is no data. 
With the error:
```  File "/Users/guy/anaconda/lib/python2.7/site-packages/artview/components/plot_radar.py", line 789, in _update_plot
    colorbar_flag=False, cmap=cmap['cmap'], mask_outside=True,```

**Comment by nguy** [comment]
oops last line ```KeyError: 'cmap'```

**Comment by nguy** [comment]
Actually here's the whole error:
```Traceback (most recent call last):
  File "/Users/guy/anaconda/lib/python2.7/site-packages/artview/components/plot_radar.py", line 510, in NewCmap
    self._update_plot()
  File "/Users/guy/anaconda/lib/python2.7/site-packages/artview/components/plot_radar.py", line 809, in _update_plot
    self._update_axes()
  File "/Users/guy/anaconda/lib/python2.7/site-packages/artview/components/plot_radar.py", line 823, in _update_axes
    self.ax.set_xlim(limits['xmin'], limits['xmax'])
TypeError: 'NoneType' object has no attribute '__getitem__'
2015-12-02 15:38:20.175 python[90466:10496655] modalSession has been exited prematurely - check for a reentrant call to endModalSession:
{'xmin': -150.0, 'ymin': -150.0, 'ymax': 150.0, 'xmax': 150.0}
  File "/Users/guy/anaconda/lib/python2.7/site-packages/artview/components/plot_radar.py", line 510, in NewCmap
    self._update_plot()
  File "/Users/guy/anaconda/lib/python2.7/site-packages/artview/components/plot_radar.py", line 789, in _update_plot
    colorbar_flag=False, cmap=cmap['cmap'], mask_outside=True,
KeyError: 'cmap'
```

gamaanderson:
from artview/components/limits.py remove lines 201 (and 202)

nguy:
Yep, that did it.

nguy:
Okay so with that change, it seems to be a good first pass at working.

nguy:
I'd like to put this out there and that way maybe get some feedback.

nguy:
Would you make the changes of deleting the lines and rename and then push the changes?

gamaanderson:
ok, just give me a couple of days, there is some cleaning up to be made.

nguy:
You bet, just put a message in here when it's ready and I'll merge.

gamaanderson:
ok

gamaanderson:
cleaning done, you may merge.
Unless you have modification you want to preserve, I would recommend to reset your branch to avoid local conflicts.

nguy:
Okay, just final testing this morning. 
I'm having trouble identifying where the problems I am having occur:

1. When starting the ```manual_unfold``` mode, ```SelectRegion_dev``` starts in menu, but does not appear to be enabled (i.e. no polygons are being drawn).
2. When using the ```SelectRegion``` through the display menu, it is using the reflectivity field instead of the velocity field which was selected for that display.

gamaanderson:
1. There is not un-enable in selectRegion, if VplotAxes is linked it must work.

2. Artview don't know the difference between reflectivity and velocity, it will always work with the field that is selected in the display

nguy:
Okay per 1, The ```SelectRegion_dev``` only works with Display1 I realize. Maybe we should consider when starting this mode naming the displays: 'VelocityUnfoldDisplay' and 'SecondaryDispaly' (or something) so it is clear which window works. I know that the user can link/unlink, but it's one less step to figure out.

nguy:
I can't reproduce the second instance.

gamaanderson:
That the down side of mode, artview will link the first display it founds, the user has no control there.

nguy:
Okay, something to keep in mind for the future if we think of a fix.
I'm happy to merge now, I think and have people give it a try.

gamaanderson:
I just made some tutorial over the three new features, just to avoid people from getting lost:

https://youtu.be/1ehZXbp7000
https://youtu.be/WuS7tvvRtiM
https://youtu.be/PYhwG8Juryo

nguy:
@gamaanderson Great work on getting this going and the video tutorials.
I should do a couple of these as well, what do you use to record your screen?

I'll probably link in a help button to these features as well to explain usage in a future update.

gamaanderson:
I used Kazam, its a KDE-screencaster, so its probably just available for linux.

nguy:
Thanks, I'll look for something in the Mac Apps.

Issue #94, opened 2015-09-15
Plugin for IO


gamaanderson:
Make plug-in for more advanced io operations. It must in any case have a tree view of directory. It could also have an optional cache of radar files.

nguy:
Something that might be nice is to pop up a list of files in directory and click on a file to display the data.

gamaanderson:
that is exactly what I mean, Qt has already specific tools for doing such.

gamaanderson:
I have this implemented in a [branch](https://github.com/gamaanderson/artview/tree/treeview). Use the view script (`artview -s view`) to give it a look. Use double clicks to move between folders (.. for back) and open files and right-click for more options.

I named this AdvancedIO, but for now it is only useful for opening. I have no good idea of how to add the write part.

nguy:
Closing issue since this was merged

PR #101, opened 2015-10-03
Treeview

gamaanderson:
This is waiting feed back (issue #94).

### DIFF
--- a/artview/plugins/io.py
+++ b/artview/plugins/io.py
@@ -1,0 +1,201 @@
+"""
+tree_view.py
+"""
+
+# Load the needed packages
+import code
+import pyart
+
+import sys
+import os
+
+path = os.path.dirname(sys.modules[__name__].__file__)
+path = os.path.join(path, '...')
+sys.path.insert(0, path)
+
+import artview
+
+from ..core import Component, Variable, common, QtGui, QtCore, componentsList
+
+open_functions = [
+    pyart.io.read,
+    pyart.io.read_grid,
+    pyart.io.read_grid_mdv,
+    pyart.aux_io.read_d3r_gcpex_nc,
+    pyart.aux_io.read_gamic,
+    pyart.aux_io.read_kazr,
+    pyart.aux_io.read_noxp_iphex_nc,
+    pyart.aux_io.read_odim_h5,
+    pyart.aux_io.read_pattern,
+    pyart.aux_io.read_radx]
+
+
+class AdvancedIO(Component):
+    '''
+    Open an interactive python console so the direct manipulation
+    '''
+
+    Vradar = None  #: see :ref:`shared_variable`
+    Vgrid = None  #: see :ref:`shared_variable`
+
+    @classmethod
+    def guiStart(self, parent=None):
+        '''Graphical interface for starting this class.'''
+        return self(), False
+
+    def __init__(self, dirIn=None, name="AdvancedIO", parent=None):
+        '''Initialize the class to create the interface.
+
+        Parameters
+        ----------
+        [Optional]
+        dirIn: string
+            Initial directory path to open.
+        name : string
+            Field Radiobutton window name.
+        parent : PyQt instance
+            Parent instance to associate to this class.
+            If None, then Qt owns, otherwise associated with parent PyQt
+            instance.
+        '''
+        super(AdvancedIO, self).__init__(name=name, parent=parent)
+        self.dirView = DirView(dirIn)
+        self.setCentralWidget(self.dirView)
+        self.dirView.opened.connect(self.open)
+        self.dirView.customContextMenuRequested.connect(self.contextMenu)
+        self.Vradar = Variable(None)
+        self.Vgrid = Variable(None)
+        self.sharedVariables = {"Vradar": None,
+                                "Vgrid": None}
+        self.connectAllVariables()
+        self.show()
+
+    def open(self, path):
+        ''' Slot of :py:attr:`DirView.opened` '''
+        # try several open
+        print ("open: %s" % path)
+        self.filename = str(path)
+        try:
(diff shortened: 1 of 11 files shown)


artview/plugins/io.py
```diff
@@ -0,0 +1,201 @@
...
+
+    Vradar = None  #: see :ref:`shared_variable`
+    Vgrid = None  #: see :ref:`shared_variable`
+
+    @classmethod
+    def guiStart(self, parent=None):
+        '''Graphical interface for starting this class.'''
+        return self(), False
+
+    def __init__(self, dirIn=None, name="AdvancedIO", parent=None):
```
nguy: This opens within the menu now, it's a little awkward on my machine. It would be nice if this window could "pop out"

artview/plugins/io.py
```diff
@@ -0,0 +1,201 @@
...
+
+        if grid_warning or radar_warning:
+            msg = "Py-ART didn't recognize this file!"
+            common.ShowWarning(msg)
+        else:
+            msg = "Could not open file, invalid mode!"
+            common.ShowWarning(msg)
+        return
+
+    def contextMenu(self, pos):
```
nguy: Is this the part that the right click brings up read options?

artview/plugins/io.py
```diff
@@ -0,0 +1,201 @@
...
+        self.Vgrid = Variable(None)
+        self.sharedVariables = {"Vradar": None,
+                                "Vgrid": None}
+        self.connectAllVariables()
+        self.show()
+
+    def open(self, path):
+        ''' Slot of :py:attr:`DirView.opened` '''
+        # try several open
+        print ("open: %s" % path)
```
nguy: This print statement might duplicate the RadarDisplay or GridDisplay message of file opening.

artview/plugins/io.py
gamaanderson: Yes, this is a component, it is just a question of changing the script to don't open in the standard start.

artview/plugins/io.py
```diff
@@ -0,0 +1,201 @@
...
+                grid_warning = True
+
+        if grid_warning or radar_warning:
+            msg = "Py-ART didn't recognize this file!"
+            common.ShowWarning(msg)
+        else:
+            msg = "Could not open file, invalid mode!"
+            common.ShowWarning(msg)
+        return
+
```
nguy: Opening a new file was unsuccessful for me.

artview/plugins/io.py
gamaanderson: yes it is, it was a little trick to make it work, but this is exactly the place. The name "context menu" is also the standard name of this kind of functionality (case you want to google it up).

artview/plugins/io.py
gamaanderson: I don't think it does, because that message is in `Menu` and this is "replacing" `Menu`.

artview/plugins/io.py
gamaanderson: really? did you got the traceback in the terminal?

artview/plugins/io.py
nguy: Good to know, that's a nice feature!

artview/plugins/io.py
nguy: Nothing in the terminal at all. Kinda stumped.

gamaanderson: My fault, I forgot to link, but if you didn't got any traceback in the terminal then it was successful. To visualize you need to link `Vradar` of `AdvancedIO` with `Display`, but this will already be in the script, so no problem.

gamaanderson: So you open it, you just didn't display it, the important part is what you think of the functionality and if there is any idea of how to add writing (from the graphical point of view)

nguy: Okay, you're right. Linked them and it worked just fine. 
I think that ```AdvancedIO``` should be changed to something more fundamental like ```FileList```.

nguy: Are you against linking the Display when this Plugin is run?

nguy: It could also be nice to add this functionality to the ```File``` Menu called something like "View Files".

gamaanderson: I'm ok with `FileList`, if you really think it is better. Please think a little and if you change your mind inform me.

Link in the script I'm ok; automatic linking at run time I'm initially against, but this should be possible (in a way I accept) in PR #102 (modes = run time scripts)

I'm initially agains adding to the `File` (because of the old code independence), but I could come to accept it if I agree to make `Menu` a non independent component which would probably involve moving it to `core` or somewhere else. It is easy to do, but it is kind of a big decision, so keep pushing me and I will give up (eventually).

nguy:
I do think FileList is better, while it actually gives the user a more advanced tool, it is easier to use and AdvancedIO may be "scary". The reason behind suggesting this tool be put into ```File``` is that it fits a bit better there and would be the obvious place I would look for this feature. I would come to learn it's in ```Advanced Tools```, but again this offers more fundamental operations on the files and not interacting with the data as many of the Advanced Tools do.

I see your point about code independence and the current structure, but we may need to make small sacrifices for usability. Hmmmm, let's chew on this for a day.

nguy:
Would you change the name to ```FileList```? I still think this is best. Let's leave it out of the File menu for now and explore this later if it makes sense. With that change I'll merge this into the master branch.

gamaanderson:
There it is, ready for merge (for now).

About the changing Menu, I think there is way to do it, but it involves some code changes, so I would like to wait a little to finish the other things first.

nguy:
Thanks @gamaanderson 
I'm merging this and we can update functionality as needed in the future.

PR #102, opened 2015-10-03
Modes

gamaanderson:
This is a proof of concept to be studied.

Artview has been created in an code independence bases, and the existing tools (GUIstart, linking and script) make it powerful for advanced user/programer. However un-advanced user are having problems with it, this adds to Menu a list of modes, clicking this modes will perform a list of action that emulate the actions of an advanced user. There are just two modes implemented, "radar" and "map_to_grid".


nguy:
I'm sorry for the slow followup I've been utterly swamped!
I'm hoping to get to some of these outstanding issues later today.

nguy:
@gamaanderson Do you have an order you'd like me to look at these different PRs?

gamaanderson:
No Problem, it is perfectly understandable, we have done a lot in the past 2 mouths. It is good to slow the rhythm a little. My classes have finally started this week, so I should also slow down.

nguy:
Oh nice, I understand. I'll start to address these one by one and we can slowly has these out.

gamaanderson:
The order is not really important, but if want one start with PR #102 and let #82 as last one

nguy:
Okay, I'm looking at #101 also.

artview/components/plot_radar.py
```diff
@@ -965,7 +966,7 @@ def setupUi(self):
         self.layout.addWidget(QtGui.QLabel("Vlims"), 3, 0)
         self.layout.addWidget(self.limsButton, 3, 1, 1, 3)
 
-        self.name = QtGui.QLineEdit("Display")
+        self.name = QtGui.QLineEdit("RadarDisplay")
```
nguy: I like the changes here and grid_radar more explicitly defining the Display used.

artview/plugins/__init__.py
```diff
@@ -28,7 +28,7 @@
         tmp = importlib.import_module('.'+module[:-3], __package__)
     for plugin in tmp._plugins:
         setattr(thismodule, plugin.__name__, plugin)
-        _plugins.append(plugin)
+        _plugins[plugin.__name__] = plugin
```
nguy: What's the advantage of dictionary vs. list here?

artview/components/menu.py
```diff
@@ -299,6 +309,49 @@ def startComponent(self, Comp):
...
+                # comp1.var = comp0.var
+                setattr(components[link[1][0]], link[1][1], dest)
+                # Connect new Variable
+                components[link[1][0]].connectSharedVariable(link[1][1])
+                # comp1.var.change(comp0.var.value), just to emit signal
+                dest.update()
+
     def addFileAdvanceMenu(self):
         '''
         Add menu to advance to next or previous file.
```
nguy: Okay, sorry did you forget to add the modes menu?

artview/plugins/__init__.py
gamaanderson: here none, in modes however I can do

    for key in _plugins.keys():
        name = key

instead of

    for plugin in _plugins:
        name = plugin._name_

yes, the diference is minimal!

artview/components/menu.py
gamaanderson: I don't think [so](https://github.com/nguy/artview/pull/102/files#diff-8105a91fbea27783d61781e9662802c9R268). I can't now, but I will test this later.

artview/plugins/__init__.py
nguy: Just curious why you went that route. Makes sense.

artview/components/menu.py
nguy: I think ```addModesMenu``` needs to be called say around L208ish.

gamaanderson: Your right, but I did test it, so I don't know, something got lost. See the new commit, this should work

nguy:
This looks pretty good. I think it's a good start for this type of functionality.

gamaanderson:
nice, that you like it, in theory it can replace most scripts.

I wounder if you got annoyed by the pop up windows? I could reduce that by making automatic decisions, but this would reduce user control. It would also happened at Component level, so or I create a new mandatory method like `startGUI` or components can't have mandatory arguments in `__init__` and Components Developers need to stick to that.

nguy:
I didn't get annoyed by the popups, but I also see their utility.
I could certainly see the argument of making some automatic decisions though to make it even cleaner interface.

It doesn't seem like tightening the guidelines for developers would cause many issues. Do we have many instances of mandatory arguments in ```__init__```?

gamaanderson:
we do have a fell mandatory arguments, but they are mostly historical and could be made optional.

nguy:
Hmm, I might vote to go that way then. What are your feelings?

gamaanderson:
Yes, I think we can go that way. Changing modes will happen by one click then even if some assumptions are made in the way. The current control "tools" shall however not be deprecated, but exist in parallel.

One disadvantage, in this one click way, the user losses control over if components are free windows or tabs in menu, which should be the standard? 
Also, One way to restore that control is with drag-drop tools (not PR #82), but this needs python >= 2.7.

nguy:
I agree that the current tools should exist in parallel and could be useful to advanced users.

Is it possible to have window tabs become freestanding windows through a right click? That might solve that problem. I like the idea of tabs as this keeps the screen clean.

The more this comes up the more I think we just say use python <2.7 at your own risk and we don't support earlier versions.

gamaanderson:
Another thing the one should mention, is that changing modes does not close anything, it just open new ones (if needed), it is up to the user to close. We could eventually add strong mode changes, that closes everything that does not belong to that mode, but this is likely to be a little too "destructive".

nguy:
I agree that closing other windows is not a good idea. The user can choose what to leave open and close.

nguy:
Do you want to leave this as proof of concept or merge in this functionality in a sort of beta form?

gamaanderson:
I prefer to left this unmerge for now.

nguy:
You got it. We'll come back to it.

PR #104, opened 2015-10-06
SelectRegion masking data

nguy:


gamaanderson:
The `edit_radar` script is not working (`_parse_dir` is not imported). It is also making absolute imports, that is not recommended.

nguy:
Darn, I didn't notice this comment until after merge. 
I'll need to look into this, and fix.

nguy:
These are addressed in PR #106

PR #113, opened 2015-12-08
Help and documentation updates.

nguy:
Made displayHelp functions internal.
Updated dialog boxes to explicitly accept HTML markup.
Added hyperlinks to modes documentation.
Updated help sections on some funtions.
Added hyperlink to documentation in Menu -> About
Updated documentation.


nguy:
If we're happy with some of the modified help functions, I'll remove the older commented out code before merging.

nguy:
I just noticed that the File -> Filelist displays all files, but is not opening the file when double-clicked. @gamaanderson can you verify on your install after last merger?

gamaanderson:
I will verify this later today, when I get home.

For the future. Please, when you build the documentation, try to do that in a separated commit, so I can see the changes to the .py files in github. Otherwise the commit is so big that I can't see it. To do that: commit changes to .py files without the "-a" flag and instead list the files you want to add to the commit.

nguy:
Yeah sorry about that, I realized this issue as I was pushing. I think all the code changes are at least listed in the top files displayed. The only one I notice isn't evident there is that I rolled rev to 1.2.1 in ```setup.py```.

gamaanderson:
Well I have just rolled to version 1.2.0 in my last commit, but if you think we should roll to 1.2.1 that is ok.

Well, I'm not sure how comfortable I'm in putting link in the source code, but I also don't have an argumentation why not to.

Beside that, for me this is good for merging.

nguy:
I went back and forth on whether to include the links. It can be viewed as a security risk for sure, but it's also only activated by clicking. @jjhelmus you are more  in tune with security stuff what do you think about including an active link in Help window?

If we had a true server to host the videos I wonder if that would be more safe?

nguy:
I've been thinking about this more and I don't see an issue with the link (as long as it remains operable). I am okay removing this in the future if we come up with a "better" (or more official) way to host the videos.

gamaanderson:
ok, as I said I don't have real reasons against it.

jjhelmus:
Not sure if you are interested but typically the build documentation is not checked into the master branch (although I have done this in the past also...).  Rather these files are more often checked into a [gh-pages branch](https://help.github.com/articles/creating-project-pages-manually/) which are then served up at user.github.io/project/.  Check out the [gh-pages branch of nmrglue ](https://github.com/jjhelmus/nmrglue/tree/gh-pages) for an example and the [resulting site GitHub generates](http://jjhelmus.github.io/nmrglue/).

nguy:
Wow, that looks nice!. I'll check this out and see if we can't figure out how to do this. 

So this is what you do for Py-ART too then?

gamaanderson:
Thank you @jjhelmus, we (at least I, and I think I commented that somewhere) knew that. I try it once, but it didn't work and I have not retried it sense then. We will need two local copies of the repository to build the doc, don't we?

jjhelmus:
No, Py-ART's documentation is a bit more complicated, probably a bit much unless updating documentation is a big pain.  Py-ART has [Travis CI](https://github.com/ARM-DOE/pyart/blob/master/.travis.yml#L50) [build the doc](https://github.com/ARM-DOE/pyart/blob/master/continuous_integration/build_docs.sh) on every commit.  When Travis build a [non-pull request from the ARM-DOE repository](http://jjhelmus.github.io/nmrglue/), typically when a PR is merged, then it creates a single commit with the new documentation and force pushes it to the gh-pages branch of the [pyart-docs-travis] (https://github.com/ARM-DOE/pyart-docs-travis) repository.  I borrowed this setup from pandas is my memory is correct.  

Another option for automatically keeping your online documentation up to date is https://readthedocs.org/, but that was not working with Py-ART's dependencies last I checked.

jjhelmus:
> Thank you @jjhelmus, we (at least I, and I think I commented that somewhere) knew that. I try it once, but it didn't work and I have not retried it sense then. We will need two local copies of the repository to build the doc, don't we?

Two local copies make things much easier.  I have two local copies of the Py-ART repository on my machine, one of which always has the gh_pages branch checked out.  

The other gh-pages hint is that you need to add a .nojekyll file in the root of the branch.  If you do not GitHub will use [Jekyll to build the content](https://help.github.com/articles/using-jekyll-with-pages/)

Issue #110, opened 2015-11-10
Py-ART aux_io 


nguy:
In ```artview.plugins.io```, we should use a logic-based approach for the enabling the ```pyart.aux_io``` read routines. 
Either a try: except: statement or test for each instance name in ```pyart.aux_io.__all__``` before loading.

jjhelmus:
Just FYI, all the aux_io read function are available in the pyart.aux_io namespace regardless of if the optional dependencies are available (ARM-DOE/pyart#411), so testing for the names will no longer work.  

You'll either need to catch the **MissingOptionalDependency** exceptions during read or examine the private testing variables in the various modules.  This is probably not ideal for ARTView but I've been trying to improve the Py-ART user experience with this change.  Also  aux_io is not intended to have full support in Py-ART so... it does not have full support.

gamaanderson:
@jjhelmus is using `__all__` a save way to get all (aux) read function (even if some will not work). In the end `__all__` is a private/special method, we probably should not use it.

jjhelmus:
`__all__` is fine to use for package (or in this case) sub-package introspection, and I believe is the preferred method..  The convention (maybe standard?) in Python is that `__all__` defines all the public names in the package or sub-package.  It is also used to determine what names get imported when importing * from a package, eg  `from pyart.aux_io import *`. 

In general names that begin and end with two underscores, `__variable__` reflect a builtin method of variable.  Other examples include `__name__`, `__file__` and `__doc__`.  

Variables with a single leading underscore are private variables but can still be access with their given names.  Two leading underscores with a single or no trailing underscores also indicates a private variable which may also be [name mangled](https://docs.python.org/2/tutorial/classes.html#private-variables-and-class-local-references) in classes to further protect them from unintended access.

gamaanderson:
using `__all__` don't work, it also list modules and I just want read functions.

jjhelmus:
Interesting.  I did not know that `from .module import func` adds module to the symbol table.  Apparently this is due to how imports work in Python (PEP 302).  The modules in pyart.aux_io could be deleted from the pyart.aux_io symbol table or `__all__` could be but explicitly defined with the function names so that only the functions/class are listed.  Apparently both of these are considered a bad practices because the modules are objects in that namespace.  Keeping the modules in there also makes debugging and testing easier.  Without them `pyart.aux_io.gamic_hdf5.read_gamic` would raise a AttributeError.

jjhelmus:
[...]

Issue #109, opened 2015-11-10
Vcmap color table value is an object or not


pfhein:
Vcmap changes depending on whether the color table.  For the color table `jet` if you `print dir(Vcmap.value['cmap'])` you will get 
```{'vmax': 65, 'cmap': <matplotlib.colors.LinearSegmentedColormap object at 0x7fe02770eed0>, 'vmin': -10}```
And you need to use `Vcmap.value['cmap'].name` to get the value `jet`.
However for color table `pyart_NWSRef` you will get
```{'vmax': 65, 'cmap': 'pyart_NWSRef', 'vmin': -10}```
So in a pyart color table  `Vcmap.value['cmap']` is not an object and it gives you  color table value `pyart_NWSRef`.  There should not be this difference.

**Comment by nguy** [comment]
This is interesting. We'll see what we can hunt down. Does this happen whenever you select a new color table or just with this color table or in general?

**Comment by pfhein** [comment]
To learn how to make a plugin, I thought I would create a simple plugin that reports the color table.  Right now it works only at plugin startup and then only with the Display 1 Window.  So it reports the current color table at plugin startup.

**Comment by pfhein** [comment]
Another thing is that when I start up artview  field variable `CZ ` is given the color table `jet` but if I chose a color table `jet` is not an option and only pyart color table values are options.  So maybe the problem is that  color table `jet` is being chosen instead of a pyart color table value.

**Comment by nguy** [comment]
Gotcha. Okay, I'm looking through the code now to see exactly where these are pulled. Can you paste your code to see what is going on? 

It's weird that ```CZ``` is not bringing up bringing up the ```pyart_NWSRef```. There is a check in artview.scripts._parse_field that should catch CZ as a valid reflectivity.

**Comment by nguy** [comment]
I'd like to see how 'jet' is being loaded, because when a ```RadarDisplay``` is started, the Py-ART colortables are loaded and supercede the matplotlib currently. When they are loaded, the string name is saved to the Vcmap.value['cmap'] field.

**Comment by pfhein** [comment]
Artview has always loaded the same color table `jet` for `CZ`.  I think it is `jet` as it does not match any of the pyart color tables.  Or is it pyart overriding artview and not recognizing `CZ`.  Here is my plugin code.  Note I copied from your plugin and it is not done.  I think it needs a button to run it properly.
```
"""
cmap_value.py
"""

# Load the needed packages
from PyQt4 import QtGui, QtCore
import pyart
from .. import core, components

[...]

Issue #108, opened 2015-11-03
Py-ART Grid refactor. 


jjhelmus:
I'm planning on a large refactor and reorganization of the **Grid** class in Py-ART and want to check on how much ARTView depends on the organization of the class.  If needed it should be possible to keep a compatibility layer which allow access to the old **Grid** attributes but if this is not needed I would like to avoid it.

nguy:
I think @gamaanderson can answer better, but it looks like axes information is the most used part of the **Grid** class in ARTView. What kind of changes are you considering?

We could probably write a check upon a **Vgrid** initialization that checks attributes to account for changes.

gamaanderson:
There is no need for a compatibility layer, we can copy the current version of `artview/components/plot_grid.py` to a new file and check pyart version while loading. All other uses of grid outside this file should not be dependent of any internal organization of grid (**thanks to artview's code independence model!**). Just inform us later which is the version to check for.

jjhelmus:
Good to hear.

I'll outline the new layout I have in mind for the **Grid** object in ARM-DOE/pyart#285 tomorrow to gather feedback before implementation.  I have it drawn out on paper but have not had the time to write it up nicely.

Issue #96, opened 2015-09-23
User Communication


gamaanderson:
So the good news is that we have a small, but growing user community. The bad news is that is community is silent. This is absolutely normal, but it would be really important to create a conversation. Ideas?

gamaanderson:
So as our components are already creating some useless outputs and I don't think we will clean that any time soon, we could take the opportunity and add in the end of every script a print asking for feed back, I don't think this will annoy any one . We could also open a mail list.

nguy:
Yes, engaging can be difficult. I'm open to a statement print out. If it's in the terminal screen or text in the Menu Window it would be very unobtrusive.

PR #116, opened 2016-01-13
Minor Updates

gamaanderson:
Updates to address issues #110, #108, #96 and test for dependencies in `setup.py` (#109)

### DIFF
--- a/artview/components/plot_grid.py
+++ b/artview/components/plot_grid.py
@@ -574,7 +574,10 @@ class GridDisplay(Component):
 
     def toolDefaultCmd(self):
         '''Restore the Display defaults.'''
-        from . import tools
+        for key in self.tools.keys():
+            if self.tools[key] is not None:
+                self.tools[key].disconnect()
+                self.tools[key] = None
         self._set_default_limits()
         self._set_default_cmap()
 
--- a/artview/components/plot_radar.py
+++ b/artview/components/plot_radar.py
@@ -602,11 +602,16 @@ class RadarDisplay(Component):
 
     def toolValueClickCmd(self):
         '''Creates and connects to Point-and-click value retrieval'''
-        from .tools import ValueClick
-        self.tools['valueclick'] = ValueClick(
-            self.Vradar, self.Vtilt, self.Vfield,
-            self.units, self.ax, self.statusbar, parent=self.parent)
-        self.tools['valueclick'].connect()
+        if self.plot_type == 'radarPpi':
+            from .tools import ValueClick
+            self.tools['valueclick'] = ValueClick(
+                self.Vradar, self.Vtilt, self.Vfield,
+                self.units, self.ax, self.statusbar, parent=self.parent)
+            self.tools['valueclick'].connect()
+        else:
+            import warnings
+            warnings.warn("valueclick not implemented for airborne or rhi",
+                          NotImplementedError)
 
     def toolSelectRegionCmd(self):
         '''Creates and connects to Region of Interest instance'''
@@ -622,11 +627,12 @@ class RadarDisplay(Component):
 
     def toolDefaultCmd(self):
         '''Restore the Display defaults.'''
-        from . import tools
-        self.tools, limits, cmap = tools.restore_default_display(
-            self.tools, self.Vfield.value, self.plot_type)
-        self.Vcmap.change(cmap)
-        self.Vlims.change(limits)
+        for key in self.tools.keys():
+            if self.tools[key] is not None:
+                self.tools[key].disconnect()
+                self.tools[key] = None
+        self._set_default_cmap()
+        self._set_default_limits()
 
     def getPathInteriorValues(self, paths):
         '''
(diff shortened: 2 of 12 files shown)


artview/components/plot_grid.py
```diff
@@ -254,11 +254,11 @@ def _fillLevelBox(self):
...
         # Loop through and create each level button
         if self.plot_type == "gridZ":
-            levels = self.Vgrid.value.axes['z_disp']['data']
+            levels = self.Vgrid.value.z['data']
         elif self.plot_type == "gridY":
-            levels = self.Vgrid.value.axes['y_disp']['data']
+            levels = self.Vgrid.value.y['data']
         elif self.plot_type == "gridX":
-            levels = self.Vgrid.value.axes['x_disp']['data']
+            levels = self.Vgrid.value.x['data']
```
nguy: I haven't followed the GridDisplay updates closely. Is this change backwards compatible? Meaning does ```self.Vgrid.value.z``` map to ```self.Vgrid.value.axes['z_disp']``` in PyART or somewhere in ARTView?

nguy: Wait now looking further below, either plot_grid or plot_grid_legacy is called within ARTView depending on version number, correct?

artview/components/plot_radar.py
```diff
@@ -602,11 +602,16 @@ def toolZoomPanCmd(self):
...
+        if self.plot_type == 'radarPpi':
+            from .tools import ValueClick
+            self.tools['valueclick'] = ValueClick(
+                self.Vradar, self.Vtilt, self.Vfield,
+                self.units, self.ax, self.statusbar, parent=self.parent)
+            self.tools['valueclick'].connect()
+        else:
+            import warnings
+            warnings.warn("valueclick not implemented for airborne or rhi",
+                          NotImplementedError)
```
nguy: I don't recall ValueClick not working with RHI/Airborne. Maybe I never tested....

artview/components/tools.py
```diff
@@ -58,6 +60,8 @@ def reset_tools(tooldict):
     -----
     Disconnects all tools.
     '''
+    import warnings
+    warnings.warn("this function is deprecated")
     for tool in tooldict:
         if tooldict[tool] is not None:
             tooldict[tool].disconnect()
```
nguy: Just curious: what's the reason for putting the tool disconnects in the display scripts versus in the tools.py?

artview/plugins/io.py
```diff
@@ -17,20 +17,30 @@
 
 from ..core import Component, Variable, common, QtGui, QtCore, componentsList
 
+# get list of read functions
+import inspect
+aux_read_functions = inspect.getmembers(pyart.aux_io, inspect.isfunction)
+read_functions = [
+    pyart.io.read,
+    pyart.io.read_grid,
+    pyart.io.read_grid_mdv] + [a[1] for a in aux_read_functions]
```
nguy: This is pretty slick! I've never used inspect. Is it a built in Python library?

artview/plugins/io.py
```diff
@@ -17,20 +17,30 @@
...
+try:
+    for func in read_functions:
+        try:
+            func(None)
+        except pyart.exceptions.MissingOptionalDependency:
+            brocken_read_functions.append(func)
+        except:
+            pass
+except:
+    pass
```
nguy: Have you run into trouble with broken functions in Py-ART?

artview/components/plot_grid.py
gamaanderson: yes!

artview/components/tools.py
gamaanderson: this is just a warning that this function is no longer being used.

I made however the opposite question, why remove from display class control over an internal part of it. There is no abstract concept of tool, it is an empiric construction, therefore any code separation is only visual.

artview/plugins/io.py
gamaanderson: `MissingOptionalDependency`

artview/components/tools.py
nguy: Fair enough. I have no problems with deprecating this. And you are right, this probably should be internal rather calling out to another location.

artview/plugins/io.py
nguy: Ah, like h5py or such. Got it, nice!

artview/plugins/io.py
gamaanderson: Yes, it is python built in. This approach has suggested by @jjhelmus

nguy:
So, I'm unable to open the gridded MDV file in the PyART testing directory. Would you mind sharing a gridded file with me to look at and make sure I can open?

gamaanderson:
The pyart example should work, but they do not open with `read_grid`, only with `read_grid_mdv`. You need to use the open options of `FileList`. If it still don't work, I will pass you one of my files.

nguy:
I tried that and nothing happens when selected ```read_grid_mdv``` for me.

nguy:
Thanks for the file. I'm ready to merge unless you have anything else.

gamaanderson:
No, that is all for this PR.
The following issues can be closed as they are solved or out-dated: #74 #79 #80 #96 #99 #108 #109 #110 (I can't close because I don't have the rights)

nguy:
Thanks for the great work @gamaanderson !

PR #123, opened 2016-01-27
BUG: GateFilter plugin works on ndarray fields

### LINKED ISSUE #121: GateFilter bug
Not every Py-ART Radar object features fields with masks. They may just be standard ndarrays. I'm not sure what controls this, but in my experience I've had to design workarounds for when this occurs. Thus, if you request the mask attribute from such a field (such as with a GateFilter), you will get an AttributeError, like so:

  File "//anaconda/lib/python2.7/site-packages/artview/plugins/gatefilter.py", line 378, in apply_filters
    self.Vradar.value.fields[field]['data'].mask)
AttributeError: 'numpy.ndarray' object has no attribute 'mask'

ARTview needs to be able to account for this, either with better exception handling or by constructing the mask from scratch. Below is an example of how I've worked around this issue in [PyTDA](https://github.com/nasa/PyTDA). With this function, I am attempting to pull standard ndarrays from Py-ART fields. However, in addition to just pulling the data out of a masked array, I need to deal with occasions when there is no mask, or if there is and there is no _FillValue dictionary item:

<pre>
def get_sweep_data(radar, field_name, sweep_number):
    # Check if _FillValue exists
    try:
        fill_value = radar.fields[field_name]['_FillValue']
    except KeyError:
        fill_value = BAD_DATA_VAL
    # Check if masked array
    try:
        return radar.fields[field_name]['data'][
            radar.sweep_start_ray_index['data'][sweep_number]:
            radar.sweep_end_ray_index['data'][sweep_number]+1][:].filled(
                fill_value=fill_value)
    except AttributeError:
        return radar.fields[field_name]['data'][
            radar.sweep_start_ray_index['data'][sweep_number]:
            radar.sweep_end_ray_index['data'][sweep_number]+1][:]
</pre>

jjhelmus:
The GateFilter plugin works on radar volumes which contain fields where the
data is stored as an ndarray not a MaskedArray object.

closes #121

### DIFF
--- a/artview/plugins/manual_filter.py
+++ b/artview/plugins/manual_filter.py
@@ -160,7 +160,7 @@ class ManualFilter(Component):
         mask_range = self.Vpoints.value.axes['range_index']['data'][:]
 
         data = np.ma.array(
-            self.Vradar.value.fields[self.Vfield.value]['data']
+            self.Vradar.value.fields[self.Vfield.value]['data'])
         data.mask[mask_ray, mask_range] = True
         self.Vradar.value.fields[self.Vfield.value]['data'] = data
 
--- a/artview/plugins/manual_filter.py
+++ b/artview/plugins/manual_filter.py
@@ -159,9 +159,8 @@ class ManualFilter(Component):
         mask_ray = self.Vpoints.value.axes['ray_index']['data'][:]
         mask_range = self.Vpoints.value.axes['range_index']['data'][:]
 
-        data = self.Vradar.value.fields[self.Vfield.value]['data']
-
-        np.ma.array(data)
+        data = np.ma.array(
+            self.Vradar.value.fields[self.Vfield.value]['data']
         data.mask[mask_ray, mask_range] = True
         self.Vradar.value.fields[self.Vfield.value]['data'] = data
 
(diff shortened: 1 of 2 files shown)


jjhelmus:
[A NetCDF file](https://drive.google.com/open?id=0BzCZObue1SeEZDJfbTBUZm5vNEk) which when read with Py-ART has a field whose data is an ndarray useful for testing.  You will need to zoom in to see the data as the spatial extent of this data is ~ 1 km by 1 km.

gamaanderson:
I think gatefilter.py is the only file I think this happens, but I'm not sure. Did you run a search to try to find it? If yes, then this would be good to merge @nguy.

Thank you @jjhelmus

jjhelmus:
I searched through the code and found one other use of the mask attribute that might cause an issue.  Fixed in 89f73a2.  This should be good to go.

gamaanderson:
I agree this is good to go. Please merge @nguy.

artview/plugins/manual_filter.py
```diff
@@ -159,9 +159,8 @@ def removeFromField(self):
         mask_ray = self.Vpoints.value.axes['ray_index']['data'][:]
         mask_range = self.Vpoints.value.axes['range_index']['data'][:]
 
-        data = self.Vradar.value.fields[self.Vfield.value]['data']
-
-        np.ma.array(data)
+        data = np.ma.array(
+            self.Vradar.value.fields[self.Vfield.value]['data']
```
nguy: Missing a parentheses at the then of L163.

nguy:
@jjhelmus this looks like a nice fix!

jjhelmus:
Good catch @nguy, added the missing parenthesis.

nguy:
@jjhelmus Great work!
Any chance I could get you to change L373 from:
```common.ShowWarning("Radar is None, can not perform filtering.")```
to 
```common.ShowWarning("Radar is None, cannot perform filtering.")```?

I'll merge then.

jjhelmus:
Done

nguy:
Awesome, thanks again @jjhelmus ! Merging.

PR #119, opened 2016-01-21
Modify Menu layout for easier user interface.

nguy:

PR #125, opened 2016-02-01
correct handel of Vfilelist

gamaanderson:
Af first I thought it would be just a simple correction, for instance adding NewFilelist to the SharedVariables dict. But I notice there were several small things that needed correction. I will make comments and hope you can figure out were it got it wrong.

### DIFF
--- a/artview/components/nav.py
+++ b/artview/components/nav.py
@@ -63,18 +63,17 @@ class FileNavigator(Component):
         else:
             self.Vgrid = Vgrid
         if Vfilelist is None:
-            self.Vfilelist = Variable(0)
+            self.Vfilelist = Variable([])
         else:
             self.Vfilelist = Vfilelist
+        self.filename = ""
+        self.fileindex = 0
 
-        self.sharedVariables = {"Vradar": None,
-                                "Vgrid": None,
-                                "Vfilelist": None}
+        self.sharedVariables = {"Vradar": self.NewFile,
+                                "Vgrid": self.NewFile,
+                                "Vfilelist": self.NewFilelist}
         # Connect the components
         self.connectAllVariables()
-
-        # Initialize variables
-        self.get_variables()
 
         # Set up the Display layout
         self.generalLayout = QtGui.QVBoxLayout()
@@ -83,6 +82,9 @@ class FileNavigator(Component):
         self.generalLayout.addWidget(self.createHelpUI())
 
         self.layout.addLayout(self.generalLayout, 0, 0, 1, 2)
+
+        self.NewFile(self.Vradar, None, True)
+        self.NewFilelist(self.Vfilelist, None, True)
 
         self.show()
 
@@ -133,19 +135,19 @@ class FileNavigator(Component):
         pixlast = QtGui.QPixmap(os.sep.join([parentdir, 'icons',"arrow_go_last_icon.png"]))
         self.act_first = self.navtoolbar.addAction(
             QtGui.QIcon(pixfirst),
-            "First file: %s"%(os.path.basename(self.Vfilelist.value[0])),
+            "First file:",
             self.goto_first_file)
         self.act_prev = self.navtoolbar.addAction(
             QtGui.QIcon(pixprev),
-            "Previous file: %s"%(os.path.basename(self.Vfilelist.value[self.fileindex - 1])),
+            "Previous file:",
             self.goto_prev_file)
         self.act_next = self.navtoolbar.addAction(
             QtGui.QIcon(pixnext),
-            "Next file: %s"%(os.path.basename(self.Vfilelist.value[self.fileindex + 1])),
+            "Next file:",
             self.goto_next_file)
         self.act_last = self.navtoolbar.addAction(
             QtGui.QIcon(pixlast),
-            "Last file: %s"%(os.path.basename(self.Vfilelist.value[-1])),
+            "Last file:",
             self.goto_last_file)
 
         gBox_layout.addWidget(self.navtoolbar)
@@ -158,13 +160,11 @@ class FileNavigator(Component):
         groupBox = QtGui.QGroupBox("File Information")
         gBox_layout = QtGui.QGridLayout()
 
-        self.dirIn = os.path.dirname(self.Vradar.value.filename)
-        self.infodir = QtGui.QLabel("Directory: %s"%(self.dirIn))
+        self.infodir = QtGui.QLabel("Directory:")
         self.infodir.setStyleSheet('font: italic 12px')
         gBox_layout.addWidget(self.infodir, 0, 0, 1, 1)
 
-        self.infofile = QtGui.QLabel("File: %s"%(
-            os.path.basename(self.Vradar.value.filename)))
+        self.infofile = QtGui.QLabel("File:")
         self.infofile.setStyleSheet('font: italic 12px')
         gBox_layout.addWidget(self.infofile, 1, 0, 1, 1)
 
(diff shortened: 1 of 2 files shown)


artview/components/nav.py
```diff
@@ -63,19 +63,18 @@ def __init__(self, Vradar=None, Vgrid=None, Vfilelist=None,
...
             self.Vfilelist = Vfilelist
+        self.filename = ""
+        self.fileindex = 0
 
-        self.sharedVariables = {"Vradar": None,
-                                "Vgrid": None,
-                                "Vfilelist": None}
+        self.sharedVariables = {"Vradar": self.NewFile,
+                                "Vgrid": self.NewFile,
+                                "Vfilelist": self.NewFilelist}
```
gamaanderson: must map `NewFilelist` here or it will have no effect. Also rename removing the capital L for consistence with the variable name

artview/components/nav.py
```diff
@@ -63,19 +63,18 @@ def __init__(self, Vradar=None, Vgrid=None, Vfilelist=None,
...
+            self.Vfilelist = Variable([])
         else:
             self.Vfilelist = Vfilelist
+        self.filename = ""
+        self.fileindex = 0
 
-        self.sharedVariables = {"Vradar": None,
-                                "Vgrid": None,
-                                "Vfilelist": None}
+        self.sharedVariables = {"Vradar": self.NewFile,
```
gamaanderson: we display the "current file", so need to update it. (I will make more comment on that in the end)

Also notice, this is our first example of a `New*` function that work for several variables.

artview/components/nav.py
```diff
@@ -84,6 +83,9 @@ def __init__(self, Vradar=None, Vgrid=None, Vfilelist=None,
 
         self.layout.addLayout(self.generalLayout, 0, 0, 1, 2)
 
+        self.NewFile(self.Vradar, None, True)
+        self.NewFilelist(self.Vfilelist, None, True)
+
```
gamaanderson: to avoid making "compatibility" test in the creation of "display", I just create them empty and force an update. This has no significant overhead!

artview/components/nav.py
```diff
@@ -194,54 +194,62 @@ def createHelpUI(self):
     #   Update Methods   #
     ######################
 
-    def _update_InfoUI(self):
-        '''Update the info label.'''
```
gamaanderson: I don't need this function any more, what it did is now done be `NewFile`

artview/components/nav.py
```diff
@@ -194,54 +194,62 @@ def createHelpUI(self):
...
-            os.path.basename(self.Vradar.value.filename)))
+#    def _update_InfoUI(self):
+#        '''Update the info label.'''
+#        self.dirIn = os.path.dirname(self.Vradar.value.filename)
+#        self.infodir.setText("Directory: %s"%(self.dirIn))
+#        self.infofile.setText("File: %s"%(
+#            os.path.basename(self.Vradar.value.filename)))
 
     def _update_tools(self):
-        '''Update the info label.'''
```
gamaanderson: This function is now much bigger, since I need to add test to handle special cases (empty dir, first file in dir etc.)

artview/components/nav.py
```diff
@@ -194,54 +194,62 @@ def createHelpUI(self):
...
-            'Next file: %s'%os.path.basename(self.Vfilelist.value[self.fileindex + 1]))
 
-    #########################
-    #   Selection Methods   #
-    #########################
+        print self.fileindex
+        print filelist
 
-    def get_variables(self):
-        '''Initialize variables.'''
```
gamaanderson: Well, this function was just a no go!

artview/components/nav.py
```diff
@@ -258,40 +266,91 @@ def AdvanceFileSelect(self, findex):
             return
         self.fileindex = findex
         self.filename = self.Vfilelist.value[findex]
-        self.menu._openfile(filename=self.filename)
```
gamaanderson: If we have redundant functionalist, we will have redundant code (I added a copy of this function in the end of the file). We shall not use code of an other component, and it is also better not, since there are normally some small differences.

artview/components/nav.py
```diff
@@ -258,40 +266,91 @@ def AdvanceFileSelect(self, findex):
...
 
     def goto_last_file(self):
         self.fileindex = len(self.Vfilelist.value) - 1
         self.AdvanceFileSelect(self.fileindex)
+        self._update_tools()
 
     def goto_prev_file(self):
         self.fileindex = self.fileindex - 1
-        self.act_prev.setToolTip(
-            'Previous file: %s'%os.path.basename(self.Vfilelist.value[self.fileindex - 1]))
```
gamaanderson: this need an index test, to avoid that just let update_tool do the work.

artview/components/nav.py
```diff
@@ -258,40 +266,91 @@ def AdvanceFileSelect(self, findex):
...
-        self.act_next.setToolTip(
-            'Next file: %s'%os.path.basename(self.Vfilelist.value[self.fileindex + 1]))
         self.AdvanceFileSelect(self.fileindex)
+        self._update_tools()
 
-    def NewFileList(self, variable, value, strong):
-        '''respond to change in radar.'''
-        if self.Vfilelist.value is None:
```
gamaanderson: the case of an empty or none list is already handle in `_update_tool`.

artview/components/nav.py
```diff
@@ -258,40 +266,91 @@ def AdvanceFileSelect(self, findex):
...
+            common.ShowWarning(msg)
+        else:
+            msg = "Could not open file, invalid mode!"
+            common.ShowWarning(msg)
+        return
 
+    def NewFilelist(self, variable, value, strong):
+        '''respond to change in filelist.'''
         if strong:
-            self.Vfilelist.change(value)
```
gamaanderson: DANGER! DANGER! DANGER!

artview/components/nav.py
```diff
@@ -258,40 +266,91 @@ def AdvanceFileSelect(self, findex):
             return
         self.fileindex = findex
         self.filename = self.Vfilelist.value[findex]
-        self.menu._openfile(filename=self.filename)
-        self._update_InfoUI()
```
gamaanderson: need not to update info. Info is update when Vradar or Vgrid is changed. So opening a new file will automatically do it

artview/components/nav.py
```diff
@@ -258,40 +266,91 @@ def AdvanceFileSelect(self, findex):
...
+        '''respond to change in filelist.'''
         if strong:
-            self.Vfilelist.change(value)
             self._update_tools()
 
+    def NewFile(self, variable, value, strong):
+        '''Respond to change in a container (radar or grid).'''
+        if hasattr(variable.value, 'filename'):
+            # Update the info label.'''
+            self.filename = variable.value.filename
```
gamaanderson: we may want to also update the `self.fileindex` and then call `_update_tools`. This is relevant in the case of an external change of Vradar (for instance open through menu). This is a personal choice, I choose not to do so. 

If you want to change that, avoid duplicate calling of `_update_tools` by removing from `goto_*_file`

artview/components/nav.py
```diff
@@ -258,40 +266,91 @@ def AdvanceFileSelect(self, findex):
...
 
+    def NewFilelist(self, variable, value, strong):
+        '''respond to change in filelist.'''
         if strong:
-            self.Vfilelist.change(value)
             self._update_tools()
 
+    def NewFile(self, variable, value, strong):
+        '''Respond to change in a container (radar or grid).'''
+        if hasattr(variable.value, 'filename'):
```
gamaanderson: before using `radar.filename` always test it first!

artview/components/nav.py
nguy: Okay, I kept going in circles with this, putting it in and then removing.

artview/components/nav.py
```diff
@@ -133,19 +135,19 @@ def createNavToolbar(self):
...
         self.act_next = self.navtoolbar.addAction(
             QtGui.QIcon(pixnext),
-            "Next file: %s"%(os.path.basename(self.Vfilelist.value[self.fileindex + 1])),
+            "Next file:",
             self.goto_next_file)
         self.act_last = self.navtoolbar.addAction(
             QtGui.QIcon(pixlast),
-            "Last file: %s"%(os.path.basename(self.Vfilelist.value[-1])),
+            "Last file:",
             self.goto_last_file)
```
nguy: Okay, so you fall back to this string upon testing, e.g. first file in a directory.

artview/components/nav.py
nguy: Nice!

artview/components/nav.py
```diff
@@ -194,54 +194,62 @@ def createHelpUI(self):
...
+                "Last file: %s" %
+                os.path.basename(filelist[-1]))
         else:
-            filename = self.Vradar.value.filename
-            if filename in self.Vfilelist.value:
-                self.fileindex = self.Vfilelist.value.index(filename)
-            else:
-                self.fileindex = 0
+            self.act_first.setEnabled(False)
+            self.act_first.setToolTip("First file:")
```
nguy: Nice set of tests applied here.

artview/components/nav.py
nguy: This is no longer needed once the variables were initialized in the beginning, correct?

artview/components/nav.py
```diff
@@ -258,40 +266,91 @@ def AdvanceFileSelect(self, findex):
...
-            'Next file: %s'%os.path.basename(self.Vfilelist.value[self.fileindex + 1]))
         self.AdvanceFileSelect(self.fileindex)
+        self._update_tools()
 
-    def NewFileList(self, variable, value, strong):
-        '''respond to change in radar.'''
-        if self.Vfilelist.value is None:
+    def _openfile(self, filename=None):
+        '''Open a file via a file selection window.'''
+        print("Opening file " + filename)
```
nguy: So this is pretty much ported from Display code. Is this a safer method than attaching to the Menu?

artview/components/nav.py
```diff
@@ -258,40 +266,91 @@ def AdvanceFileSelect(self, findex):
...
+        else:
+            msg = "Could not open file, invalid mode!"
+            common.ShowWarning(msg)
+        return
 
+    def NewFilelist(self, variable, value, strong):
+        '''respond to change in filelist.'''
         if strong:
-            self.Vfilelist.change(value)
             self._update_tools()
```
nguy: Eureka!

gamaanderson:
I don't think we should have the File Information section or it should be internal (not dependent of Vradar) even if it seems strange.

This is because first displaying information is not the task of the plugin. But second and must important there is not external well defined "current file". I plugin can have an internal current (for instance `DirectoryView` has one = current select entry) but this is not necessarily the "current" open file. 

nevertheless I follow you lead, and did not remove that (I even improve it). There price for that are some very subtle points in the code, that say something is not right (the code speak if you are open to hear it!).

artview/components/nav.py
nguy: I recall you said this before, but it got lost in my code updates. Doh!

artview/components/nav.py
gamaanderson: No, I don't need this at all! All I need is in line 69

        self.filename = ""
        self.fileindex = 0

artview/components/nav.py
nguy: By just calling ```_update_tools``` the ```self.fileindex``` should be updated, no?

I wonder if we took ```_update_tools``` out of all ```goto_*``` functions and just put it in ```_openfile```?

artview/components/nav.py
gamaanderson: Menu is such that opening a new file also update the filelist. We don't need this here. It look like the same function but it is not. And also we don't want a permanent connection with menu, otherwise we would not need shared variables.

artview/components/nav.py
gamaanderson: yes, you are right.

Better in `AdvanceFileSelect`.

artview/components/nav.py
nguy: Okay, gotcha.

nguy:
Your updates are great. There's a couple of things i there I would not have figured out. I think the information is something that is useful. Let's revisit in the future if we see a way to improve or it's use grows.

nguy:
I'm going to merge this into the other PR and then we can all take one final look before merging into the master.

artview/components/nav.py
nguy: I'll make that change in the other PR code after merging this.

gamaanderson:
ok

artview/components/nav.py
gamaanderson: any way, if you call `_update_tools`, you also don't need in `AdvancedFileSelect` or `_openfile`, this will also generate duplicated calls!

artview/components/nav.py
```diff
@@ -194,54 +194,62 @@ def createHelpUI(self):
...
-        self.act_prev.setToolTip(
-            'Previous file: %s'%os.path.basename(self.Vfilelist.value[self.fileindex - 1]))
-        self.act_next.setToolTip(
-            'Next file: %s'%os.path.basename(self.Vfilelist.value[self.fileindex + 1]))
 
-    #########################
-    #   Selection Methods   #
-    #########################
+        print self.fileindex
+        print filelist
```
gamaanderson: remove this prints, they are not python3!

gamaanderson:
I'm going to sleep. I see the other PR later!

artview/components/nav.py
nguy: So see the change I just pushed to PR #119 , is this what you mean?

nguy:
Okay, thanks for your help @gamaanderson

PR #131, opened 2016-02-25
Rename SelecRegion

gamaanderson:
rename `SelectRegion` to `DisplaySelectRegion` and `SelectRegion_dev` to `SelectRegion`. Also update readme adding link to user forum.


nguy:
So this all looks good to me, it cleans up the code and makes it easier for me at least to follow in the future. 

I notice on my machine the SelectRegion can be quite slow and lose it's vertex-to-vertex lines. I think this is something for another issue. If you have no objections, I'll merge this.

gamaanderson:
I have no objections!

Issue #130, opened 2016-02-19
Menu window


nguy:
I notice on my machine, sometimes the menu window gets lost. For most people this won't be a big issue, but we may want to consider if there is a way to bring the window to the front somehow. A menu link or something?

gamaanderson:
This is a real problem, and it is also a problem in one of my inspirations (GIMP). I have to see how they are working that out.

nguy:
Potentially this? http://pyqt.sourceforge.net/Docs/PyQt4/qwidget.html#raise

nguy:
I tried the above with no luck (possibly due to my use of MacOS). Seems to be an outstanding problem and after a bit of web searching it looks to be OS dependent as well.

nguy:
Added the ```raise_()``` method to the menu component. And to test on FileNavigator. On my mac I see it momentarily come to the foreground, but then become buried again.

gamaanderson:
@nguy as this is system dependent, could you make a video showing this problem in your system? I think I know what you mean, but I would like to be sure that your experence is the same as mine.

nguy:
I shared a short video in your email.

gamaanderson:
Interesting how IOS externalize the menu bar.
If you pay attention the menu does blink for a moment in the upper corner where it should be. That mean it's IOS that is interfering in the Qt work, we have to avoid that.

nguy:
Yep, you are correct. This is why I indicated that it is an OS-dependent issue. I don't know if there is a solution for this behavior.

gamaanderson:
have you seen this page "http://doc.qt.io/qt-5/osx-issues.html"?

gamaanderson:
in script
`app.setAttribute(QtCore.Qt.AA_MacPluginApplication, True)`

nguy:
Think this could be put in at startup as a check with a conditional statement identifying MacOSX?

gamaanderson:
first of all, did it work?

nguy:
Yes, this does work to put the menu on the menu window. To implement, we'd have to install an OS check. I feel like this should be automatic in the background.

The issue remains how to bring the menu window to the front for operation.

nguy:
To test and internet search returns the following code:
```
import os
import platform

if os.name == 'posix':
    if platform.system().lower() == 'darwin':
        print("You are using MacOSX")
```

[...]

PR #136, opened 2016-03-21
STY: Override MacOSX menu placement to show in MainMenu window.

nguy:
This implements a check to see if the OS is Mac and applies an override of menu placement if it is. Mentioned in Issue #130.


nguy:
I'd like to be more consistent about micro revs when making these changes. If we decide to merge this.

gamaanderson:
when do we change the micro? we can't do it at every small PR.

nguy:
Nope, for example the airborne PR I don't think needs this change. However, since this changes the operation on the Mac side (moving menu from top to menu window), I think the operation of the package has changed and therefore the micro should be rolled. My thought is that fixes and small changes should not roll the micro version, but when the way ARTview operates changes (both to user and on developer side), then we roll numbers. Thoughts?

nguy:
@gamaanderson What do you think about this PR?

gamaanderson:
could you make that test in `view.start`?

nguy:
Added that and also to ```artview.scripts.view```. These were the 3 places I found an instance of ```app```.

gamaanderson:
ok, i'm fine with merging

nguy:
Great. What do you think about a rev roll for this?

nguy:
I'm also going to close Issue #130 for the time being. I don't think the original argument can be addressed at the moment.

gamaanderson:
I have no problem with rolling the version, but doing it before merging will create a conflict.

nguy:
Ah because of the other PR we just did.

Issue #138, opened 2016-03-21
Aspect Ratio Option


stechma2:
If possible, it would be very useful to add an aspect ratio adjustment into the display options. In particular, I work with tail radar data primarily, and the default aspect ratio produces a very "squished" display which can be difficult to interpret.

gamaanderson:
ok, that can be an easy add. Can you post one image,  so that I have an better idea on how that is showing up to you?

gamaanderson:
but before that, have you tried "adjust display limits"? I know it is not optimal, but this should be able to unsquish your image.

stechma2:
Here is the image with the default aspect ratio, and display limits set to incorporate the entire observable domain:
<img width="872" alt="screen shot 2016-03-21 at 2 06 19 pm" src="https://cloud.githubusercontent.com/assets/6628522/13931000/57d6087e-ef6f-11e5-96d3-97b4ffd69a41.png">
Here is the same image, with the same display limits, but with aspect ratio set to 1 
(using this hard coded addition to `plot_radar`): `self.display.set_aspect_ratio(aspect_ratio=1,ax=self.ax)`
<img width="872" alt="screen shot 2016-03-21 at 2 09 03 pm" src="https://cloud.githubusercontent.com/assets/6628522/13931062/9b368f94-ef6f-11e5-8e48-55c8ef848d61.png">
And finally, an image using the default aspect ratio, but adjusting the display limits and panning to unsquish and center the data. This method works, but is not ideal.
<img width="872" alt="screen shot 2016-03-21 at 2 13 35 pm" src="https://cloud.githubusercontent.com/assets/6628522/13931120/d15ffcc2-ef6f-11e5-8528-60771222423d.png">

gamaanderson:
Ok, I see the difference, I will implement right away.

gamaanderson:
I just submitted a PR for that. The aspect radio can be found in the adjust limits.

thank you again @stechma2 for helping improving ARTView.

stechma2:
Great! Thank you for doing this @gamaanderson.

nguy:
Tangentially, I finally figured out why the file does not come up with the default Airborne x/y ranges. The *scan_type* value of the radar variable is not recognizing these particular files as airborne. I have a workaround that I'll comment on in the PR. We'll have this out in the very near future with this fix.

nguy:
Fixed with merge of PR #140

PR #140, opened 2016-03-21
ENH: add aspect radio to limits dialog

gamaanderson:
enhancement according to issue #138


nguy:
In [this section](https://github.com/nguy/artview/blob/master/artview/components/plot_radar.py#L900) there is a test for airborne radar. Can you change:
```
if (radar.metadata['platform_type'] == 'aircraft_tail' or
    radar.metadata['platform_type'] == 'aircraft'):
```
to:
```
if 'aircraft' in radar.metadata['platform_type']:

File: artview/components/limits.py [hunk 1]
```diff
@@ -120,7 +120,7 @@ def _default_limits(field, plot_type):
 ###############################
```
nguy: In [L59-64](https://github.com/nguy/artview/blob/master/artview/components/limits.py#L59), we could set an aspect ratio for airborne or rhi identified files.

gamaanderson: current aspect ratio is defined as auto, I would be careful on changing that resizing may become problematic.

gamaanderson:
see new commit to the airborne change.

nguy:
The changes look good. Last thing is that I feel the other limits are pretty self-explanatory. But the aspect ratio may not be to people. Should we add a help button that pops up a window like we do in other areas?

nguy:
@stechma2 See the image below. Would you know how to use this?
screen shot 2016-03-22 at 8 17 54 am

gamaanderson:
what would be the help text. If it is not too long it could be a tool tip.

nguy:
Oh, that's a great idea! Thoughts for tooltips:
Data Min: "Minimum data value to display"
Data Max: "Maximum data value to display"
X Min: "Minimum X-axis value"
X Max: "Maximum X-axis value"
Y Min: "Minimum Y-axis value"
Y Max: "Maximum Y-axis value"
Aspect Ratio: "Y-to-X size display ratio, also accepts auto and equal"

stechma2:
@nguy I would know how to use that, though I think adding the tooltips would be very useful (I like the ones you suggested). Also, glad to see the file being recognized as airborne now with the auto-scaling applied. Thanks for these changes!

gamaanderson:
I also like the suggestion. May be we could change X and Y for "horizontal" and "vertical", as to avoid confusion with the x and y direction in grid.

gamaanderson:
just uploaded a commit with tooltips

nguy:
Nice changes. I think this is ready for merge. Agreed?

nguy:
@gamaanderson Can you make one more additional change I was just made aware of by a user?
In the README, under the install section. artview is mispelled as artivew. Since this will likely be merged soon.

gamaanderson:
done, also rolled the micro version, as this has a functionality change (was that procedure correct?)
I'm ok with merging.

Issue #126, opened 2016-02-03
new norm argument


gamaanderson:
PyART recently introduced a new argument for non-linear colormap (see [here](https://github.com/ARM-DOE/pyart/issues/390#issuecomment-178825935)). I'm not really sure how to add this to artview, I would like some ideas (@nguy,@jjhelmus).

Any way, this is just for artview 1.3. We need to finish 1.2 still.

nguy:
My first instinct would be to add a 'norm' keyword to the Vcmap. Have it default to None which uses linear Normalize() function and then add the norm=cmap['norm'] argument to the plotting routines.

To allow adjustments, I guess we could add a dialog box something like the limits on the drop-down Display Options toolbar.

gamaanderson:
and how does that new key and the current vmin and vmax work together?

Of course, adding a norm argument make little sense with out a plugin to edit the colormap, but we have being needing some improvement in this any way.

nguy:
Having a value of None should not have any effect since the vmin/vmax are already passed to the plot routines. It could be instituted in the ```artview.components.plot_radar._set_default_cmap``` and plot_grid version.

We could institute an additional step of calculating the norm, but take look [here](https://github.com/nguy/AWOT/blob/master/awot/graph/radar_horizontal.py#L173). This is a little bit different implementation, but the idea is the same. I feel this is something that could be done behind the scenes in Py-ART, by using the cmap and a keyword of boundaries. What do you think @jjhelmus ?

nguy:
I agree that this would be a nice feature for ARTview.

jjhelmus:
How to deal with the `norm` parameter in ARTView is tricky.  

My idea of adding to parameter was to allow users to create a categorical or non-linearly scaled colormap by passing a Matplotlib [BoundaryNorm](http://matplotlib.org/api/colors_api.html#matplotlib.colors.BoundaryNorm) or [LogNorm](http://matplotlib.org/api/colors_api.html#matplotlib.colors.LogNorm) instance.  This unlocks some interesting plotting possibilities but also means that the user needs to do some work to create the object to pass in as the `norm` argument. 

ARTView could allow a users to switch to a log scaling colorbar without the addition of too much additional code, the LogNorm instance would need to be created inside ARTView.  As @gamaanderson mentioned the categorical option will require more work as first some method allowing the user to specify the bounds would need to be created.  

> and how does that new key and the current vmin and vmax work together?

[...]

PR #142, opened 2016-03-31
lock colormap + radar collection

gamaanderson:
add a lock option in the colormap to avoid resetting it when changing the field (can be seen/set in the display limits dialog)
add a plugin to add and remove radar from a list of radars (`VradarCollection`). This is pretty raw and for now it is only tool for programmers. That is the reason there is no associated mode, access only in the plugins menu.


gamaanderson:
just added a new `ColormapEdit` plugin for issue #126

artview/plugins/__init__.py
```diff
@@ -31,6 +32,8 @@
         _plugins[plugin.__name__] = plugin
         # update docstring to add plugin
         __doc__ = __doc__ + """    %s\n""" % plugin.__name__
+    if hasattr(tmp,'_modes'):
+        _modes += tmp._modes
```
nguy: What is this code doing? I don't follow.

nguy:
I'm still not convinced that this is the place for the RadarCollection to exist. But I'll hold off more comment until I test the code out a little.

nguy:
After a short bout of testing, only the linear colorbar is working for me.

nguy:
On a weird plus side, the menu that we had moved to MainMenu are back. For a while my system was putting them up in the tray once again. I can't figure out what is different.

nguy:
I really like the idea of the colorbar stuff you did here.

One suggestion I have is that we add divider lines in the Modes menu like I put in the Files menu to categorize the operations.

artview/plugins/__init__.py
gamaanderson: allowing plugins to define their own modes

gamaanderson:
The colormap is working fine for me, it only display error messages when expected, for instance logarithm with negative number or non decreasing color bounds.

I have no problem with divider lines, just tell me where to put then (between which items).

artview/plugins/cmap.py
```diff
@@ -0,0 +1,307 @@
...
+    def select_cmap(self, post):
+        '''Select cmap from a list of pyart colormaps.'''
+        cmap = core.common.select_cmap().selection
+        if cmap is not None:
+            self.cmap['cmap'] = cmap
+            self.plot()
+
+    def apply(self):
+        '''Apply changes to shared colormap.'''
+        self.Vcolormap.change(self.cmap)
```
nguy: If, for example, the maximum value is changed and **Apply** is pushed before going to another cell, the modified text is not recorded. Maybe a call to ```_update_colormap``` is needed to make sure the proper values are recorded? Or possible a sub method that only gets the values from the lines?

nguy: To be clear this is an issue in many of our plugins.

nguy:
Okay, still interesting behavior.
If I choose discrete and add levels and hit apply. The colorbar does not appear.
I then go back to linear, colorbar returns and looks good.
I then choose discrete and the previous values are saved. I hit apply and the colorbar appears.

I have a screen recording if you'd like to see, but it's too large to post here.

artview/plugins/cmap.py
gamaanderson: I agree with calling `update_colormap` on apply. This may solve the issue with the discrete, It look like in your system it does not allow limits with equal values, I'm not really sure how to avoid that.

nguy:
I added this is and it seems to work. 
I think I'll merge now as it seems good with that minor fix.

Let's know that the radarcollection is there and comment further in future pull requests. Agreed?

nguy:
This looks good. Merge?

gamaanderson:
agreed, but I changed you fix, so that the update of `cmap` is called before changing `Vcolormap`. Since the update is made in place in `cmap`, I expect the exact same effect. The order diference is only important in the case of non-in-place update.

gamaanderson:
yes, this is good to merge

PR #154, opened 2016-07-06
ENH: Add Graphical Edges keyword to plots.

nguy:
Added Shared variable to allow the user to select whether to ignore edge gates, by passing the edges keyword.

### DIFF
--- a/artview/components/plot_radar.py
+++ b/artview/components/plot_radar.py
@@ -37,6 +37,7 @@ class RadarDisplay(Component):
     Vlimits = None  #: see :ref:`shared_variable`
     Vcolormap = None  #: see :ref:`shared_variable`
     Vgatefilter = None  #: see :ref:`shared_variable`
+    VignoreEdges = None #: see :ref:`shared_variable`
     VplotAxes = None  #: see :ref:`shared_variable` (no internal use)
     VpathInteriorFunc = None  #: see :ref:`shared_variable` (no internal use)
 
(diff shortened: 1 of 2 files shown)


gamaanderson:
I don't think adding a shared variable is the way to go. And most importantly, this does not give a graphical control.

nguy:
The reason I used a shared variable was to allow the setting to be retained so it doesn't have to be selected each time, like the ```Vlimits``` for example. Is there a better way you can think of?

gamaanderson:
It can be a simple variable, it does not need to be shared, unless you expect external control.

nguy:
Will the setting be retained when advancing to the next file?

gamaanderson:
of course, is a variable of the `artview.RadarDisplay` class, as long as its the same instance the setting retained.
You could re-set if you want, but only if you want and you would have to write the re-set line.

nguy:
I guess as long as it's a class variable we should be good

nguy:
Oops just realizing that as you typed. Let me monkey with it and just save it as such.

gamaanderson:
ok

gamaanderson:
ok

nguy:
Better?

gamaanderson:
yes this is good as a fast add, and I would probably end up doing exactly the same. I do still want a easier way of working with pyart arguments, but let us just merge this for now. As I said I don't think this is a problem in grid.

nguy:
Okay, I'll leave grid out for now, it can easily be added later.

I agree about the pyart arguments, but I'm not sure how to deal with different data type inputs etc. I can't think of an easy way to get these.

Issue #152, opened 2016-06-19
Despeckle


gamaanderson:
I believe [this](http://arm-doe.github.io/pyart-docs-travis/user_reference/generated/pyart.correct.despeckle_field.html#pyart.correct.despeckle_field) to be an important new addition to pyart and would like see it implement to artview. I'm kind of full for the next mouths, so if any one wants to give a try just let me know.

nguy:
Closed since @gamaanderson implemented.

PR #157, opened 2016-07-20
add despeckle

gamaanderson:
issue #152


nguy:
This is really nicely done @gamaanderson . Great addition!

I think that we need to put some help info somewhere. Especially for the ```object_field```.

gamaanderson:
just add more info to object_field and despeckle button

nguy:
That looks great. I'd say we can merge if you are okay with that.

gamaanderson:
Yes, I think that is all.

PR #158, opened 2016-07-23
add profiling

gamaanderson:
add a new attribute the version file (that is not mapped in git), which when active will dump a profile to file profile.txt when artview is run from command line. This do not affect the normal user, since the standard is not profiling and one needs to change the **version.py** file manually to change that.

I pretend to let this on in my system and once in a while give a look in there, to find optimization points.

### DIFF
--- a/artview/__main__.py
+++ b/artview/__main__.py
@@ -7,15 +7,31 @@ path = os.path.join(path, '..')
 sys.path.insert(0, path)
 
 import artview
-
+try:
+    from version import profiling
+except:
+    profiling = False
 
 def main(argv):
+    if profiling:
+        import cProfile, pstats, StringIO
+        pr = cProfile.Profile()
+        pr.enable()
+
     script, DirIn, filename, field = artview.parser.parse(argv)
 
     if script:
         artview.scripts.scripts[script](DirIn, filename, field)
     else:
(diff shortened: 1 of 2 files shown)


nguy:
This is a good idea. Should we add a command line flag maybe ```-p``` that could set the flag to true, that way we could ask a user to run this for us if needed?

nguy:
Let me clarify, I wonder if this would be a way to to find issues that people have when running the code as well as system performance?

nguy:
Okay, so I get the following error:
```
guy:artview $ python setup.py install
testing matplotlib qt backend ...
... test passed
  File "setup.py", line 219, in <module>
    setup_package()
  File "setup.py", line 159, in setup_package
    write_version_py()
  File "setup.py", line 137, in write_version_py
    'isrelease': str(ISRELEASED)})
KeyError: u'isprofiling'
```

I assume it comes from the fact that you state that it uses a non-git mapped file? 

Actually does L108 need quotes since this is a string?

gamaanderson:
I corrected that. I misunderstand how that part of the code was working, it should go now.

gamaanderson:
I do not think this is useful to find bugs, it is more for performance analysis.

PR #162, opened 2016-08-09
BUG: Fix ValueClick bug to display proper information for scan types.

nguy:
I found when using ARTview in RHI and Airborne mode that the ValueClick was not displaying the correct information. I generalized the **ValueClick** class for ```plot_radar``` similar to the ```plot_grid``` and also updated the tools to be reset when a different file type is opened.

### DIFF
--- a/artview/components/plot_radar.py
+++ b/artview/components/plot_radar.py
@@ -647,16 +647,10 @@ class RadarDisplay(Component):
 
     def toolValueClickCmd(self):
         '''Creates and connects to Point-and-click value retrieval'''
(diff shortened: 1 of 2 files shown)


gamaanderson:
It looks ok, but I can not be sure these calculation are correct just by looking at it. If you have tested, I would be ok with it.

nguy:
I tested with a few files and also compared to Solo Editor for an airborne example.
I'll merge this and move onto the next item. I'll be working on adding text to the image as a tool next.

gamaanderson:
ok

PR #165, opened 2016-10-09
ENH: Add Image Text. Modify path links for tools.

nguy:
This first PR is for comments. This was a little more difficult than I imagined, requiring sending the display to the new `ImageTextBox` tool and housekeeping was a bit of a bear to sort out.

Once we work out the details, I'll build this into `plot_grid` as well. That should be reasonably simple.

### DIFF
--- a/artview/components/display.py
+++ b/artview/components/display.py
@@ -10,7 +10,8 @@ from .. import core
 from datetime import datetime as dt
 
 
-class ImageTextBox(core.Component):
+#class ImageTextBox(core.Component):
+class ImageTextBox(QtGui.QMainWindow):
     '''
     Interface for executing :py:class:`ImageTextBox`.
     '''
@@ -28,7 +29,7 @@ class ImageTextBox(core.Component):
             Parent instance to associate to this class.
             If None, then Qt owns, otherwise associated w/ parent PyQt instance
         '''
-        super(ImageTextBox, self).__init__(name=name, parent=parent)
+        super(ImageTextBox, self).__init__(parent=parent)
         self.central_widget = QtGui.QWidget()
         self.setCentralWidget(self.central_widget)
         self.layout = QtGui.QGridLayout(self.central_widget)
@@ -67,7 +68,7 @@ class ImageTextBox(core.Component):
 
         # Add the Add Text option to top of list
         self.dispCombo.addItem('Add Text')
-        self.dispChoiceList.append(self._blank_entries())
+        self.dispChoiceList.append(self._init_entries())
 
         # Fill in the rest of the list with existing text boxes
         for tx_inst in self.display.disp_text.keys():
@@ -160,13 +161,11 @@ class ImageTextBox(core.Component):
 
     def chooseText(self, selection):
         '''Get Display Text.'''
-##        selection = self.dispCombo.currentIndex()
         self.choice_key = str(self.dispCombo.currentText())
         self.choice = self.dispChoiceList[selection]
-        self._rebuild_entry()
+##        self._rebuild_entry()
 
         self.dispCombo.setCurrentIndex(selection)
-##        print(selection)
 
     def _displayHelp(self):
         '''Display help.'''
@@ -174,9 +173,10 @@ class ImageTextBox(core.Component):
             "<b>Using the ImageTextBox window</b><br><br>"
             "<i>Choose an existing Text box:</i><br>"
             "  1. Select instance from drop-down list.<br>"
-            "  2. Modify exisint instance by changing values in "
+            "  2. Modify exising instance by changing values in "
             "     Text Value and Position. Click Update to save.<br>"
             "  3. Create a new Text box by choosing Add Text from drop-down."
+            "     Modify default values provided. Click Update to save."
             "<br>"
             "  4. Delete an instance by selecting and click Delete Text.<br>"
             "  5. Clear all instances by clicking Clear All Text.<br>"
@@ -213,6 +213,7 @@ class ImageTextBox(core.Component):
             self.dispChoiceList[select] = self.choice
             self.dispCombo.setCurrentIndex(select)
         # Write the updated text item to the Display Text instance
+        ## NG Modify to use a more unique identifier than text ##
         self.display.disp_text[self.choice['text']] = self.choice
 
         # Redraw the canvas to place new text
@@ -231,14 +232,14 @@ class ImageTextBox(core.Component):
             self.dispCombo.removeItem(delselect)
             self.display.fig.canvas.draw()
 
-        self.choice = self._blank_entries()
+        self.choice = self._init_entries()
         self.dispCombo.setCurrentIndex(0)
         self._rebuild_entry()
 
     def clrDispText(self):
         '''Clear all Display text boxes.'''
         self.display.disp_text = {}
-        self.choice = self._blank_entries()
+        self.choice = self._init_entries()
         # Delete each entry, need to reverse indices
         for i in range(self.dispCombo.count())[::-1]:
             if i > 0:
@@ -247,8 +248,7 @@ class ImageTextBox(core.Component):
                 self.dispCombo.removeItem(i)
         self.display.fig.canvas.draw()
         self.dispCombo.setCurrentIndex(0)
-        self._rebuild_entry()
-##        self.display._update_plot()##NG
+##        self._rebuild_entry()
 
     #####################
     #   Entry Methods   #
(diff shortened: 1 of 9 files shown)


gamaanderson:
I see no major changes, have you upload all files?

nguy:
Oh man, sorry. I forgot to add the main file to the commit! Should be in there now.

gamaanderson:
Ok, now it is there. I plan to see this in the weekend

nguy:
Great, thanks @gamaanderson

artview/components/display.py
```diff
@@ -0,0 +1,297 @@
+"""
+display.py
```
gamaanderson: what is the purpose of this file? the name is to general that may mislead a code reader.

artview/components/plot_grid.py
```diff
@@ -586,7 +586,7 @@ def cmapSelectCmd(self, cm_name):
 
     def toolZoomPanCmd(self):
         '''Creates and connects to a Zoom/Pan instance.'''
-        from .tools import ZoomPan
+        from .toolbox import ZoomPan
```
gamaanderson: I don't understand the need of changing tools to toolbox, and when there is not good reason for change, maintaining things as they are is a reason to not do it.

artview/components/plot_radar.py
```diff
@@ -311,6 +314,11 @@ def _units_input(self):
             self.units = val
             self._update_plot()
 
+    def _add_ImageText(self):
+        '''Add a text box to display.'''
+        from .display import ImageTextBox
+        itext = ImageTextBox(self)
```
gamaanderson: you need to save this variable to avoid garbage collector

artview/components/display.py
```diff
@@ -0,0 +1,297 @@
...
+"""
+
+# Load the needed packages
+from ..core import QtGui, QtCore
+from .. import core
+from datetime import datetime as dt
+
+
+class ImageTextBox(core.Component):
+    '''
```
gamaanderson: This is not a real component, therefore it should not be a class instance of it. I believe a simple `QMainWindow` should be enough

artview/components/display.py
```diff
@@ -0,0 +1,297 @@
...
+        self.ent_tex.setText(str(self.choice['text']))
+        self.ent_xpos.setText(str(self.choice['xpos']))
+        self.ent_ypos.setText(str(self.choice['ypos']))
+        self.ent_texc.setText(str(self.choice['col']))
+        self.ent_texsz.setText(str(self.choice['size']))
+        self.ent_texst.setText(str(self.choice['style']))
+
+    def _blank_entries(self):
+        '''Initialize Add Text Instance.'''
+        blank_text = {'text': '', 'xpos': '', 'ypos': '',
```
gamaanderson: start this value with examples, also add the options in the help

artview/components/display.py
```diff
@@ -0,0 +1,297 @@
...
+    #########################
+
+    def chooseText(self, selection):
+        '''Get Display Text.'''
+##        selection = self.dispCombo.currentIndex()
+        self.choice_key = str(self.dispCombo.currentText())
+        self.choice = self.dispChoiceList[selection]
+        self._rebuild_entry()
+
+        self.dispCombo.setCurrentIndex(selection)
```
gamaanderson: this line is redundant.

artview/components/display.py
```diff
@@ -0,0 +1,297 @@
...
+            except:
+                print("No instance found")
+            self.choice['instance'] = self.display.ax.text(
+                self.choice['xpos'], self.choice['ypos'],
+                self.choice['text'], style=self.choice['style'],
+                color=self.choice['col'], fontsize=self.choice['size'])
+            self.dispChoiceList[select] = self.choice
+            self.dispCombo.setCurrentIndex(select)
+        # Write the updated text item to the Display Text instance
+        self.display.disp_text[self.choice['text']] = self.choice
```
gamaanderson: you can not use the text as key as it is not unique.

artview/components/display.py
```diff
@@ -0,0 +1,297 @@
...
+        self.choice = self._blank_entries()
+        # Delete each entry, need to reverse indices
+        for i in range(self.dispCombo.count())[::-1]:
+            if i > 0:
+                self.dispChoiceList[i]['instance'].remove()
+                del self.dispChoiceList[i]
+                self.dispCombo.removeItem(i)
+        self.display.fig.canvas.draw()
+        self.dispCombo.setCurrentIndex(0)
+        self._rebuild_entry()
```
gamaanderson: calling `_rebuild_entry` is redundant after `setCurrentIndex`

artview/components/display.py
```diff
@@ -0,0 +1,297 @@
...
+from datetime import datetime as dt
+
+
+class ImageTextBox(core.Component):
+    '''
+    Interface for executing :py:class:`ImageTextBox`.
+    '''
+
+    def __init__(self, display,
+                 name="ImageTextBox", parent=None):
```
gamaanderson: parent must be display it self, to ensure proper closing.

artview/components/display.py
```diff
@@ -0,0 +1,297 @@
...
+
+    def _blank_entries(self):
+        '''Initialize Add Text Instance.'''
+        blank_text = {'text': '', 'xpos': '', 'ypos': '',
+                      'col': '', 'size': '', 'style': ''}
+        return blank_text
+
+    def _check_entries(self):
+        '''Check that entries are valid.'''
+        #NG Put in warnings?
```
gamaanderson: make some kind of graphical warning, text in the terminal is likely to be over seen

artview/components/plot_radar.py
nguy: Should be in next push.

artview/components/plot_grid.py
nguy: The reason I did this was to avoid confusion with the tools dictionary that is used. I could revert back if you like.

artview/components/display.py
nguy: Have default values now.

artview/components/display.py
nguy: Removed in next push.

artview/components/plot_grid.py
gamaanderson: if the only use is a `from * import`, there is not risk of confusion

artview/components/display.py
nguy: Alright, put the messages through `core.common.ShowWarning`

artview/components/plot_grid.py
nguy: I see your point, but I'd like to keep this change and move it forward.

artview/components/display.py
nguy: I was intending this to overwrite instances if changes are desired, that way the list does not continue to grow and also updates can occur to plot.

artview/components/display.py
nguy: Done

artview/components/plot_grid.py
gamaanderson: ok

artview/components/display.py
gamaanderson: this behavior will create uncontrollable and unintuitive behavior, for instance it does not allow two entries of the same text (in different positions).

artview/components/display.py
gamaanderson: NO, you need  `_rebuild_entry()` here, otherwise the entries are not updated by changing index.
`setCurrentIndex` is redundant, because it will call `chooseText` again. Not sure why there is not infinity loop here.

artview/components/display.py
gamaanderson: This is being set, but is not appearing, you must update the entries after building them.

artview/components/display.py
gamaanderson: I know this is kind of unimportant, but I really think the name **display.py** does not reflect the file content.

nguy:
If you agree with the naming, I'll merge so we can move forward again. Sorry for the delay and thanks for your comments and patience!

gamaanderson:
Yes, I'm ok with the naming. But there are still issues with `_rebuild_entry` and `_init_entries` (as comment 2 day ago). Make sure you also get those. If you have not the time to work on that, I can make the code alteration my self.

nguy:
I think the latest commits address everything. If anything is outstanding let me know and I'll fix it up.

gamaanderson:
**plot_radar.py** in line 319 still importing from display, not image_text, otherwise this is good to merge

nguy:
Thanks, that did it!

PR #169, opened 2016-11-16
Update image_text.py

gamaanderson:
We were kind of talking pass each other. So I implemented the changes my self, may be now it is clear what the problem was.


nguy:
Thanks, I'll merge this and hopefully get this thing finished!

Issue #166, opened 2016-10-11
Tilt information update


nguy:
When the tilt is changed (up-down arrows or selection), the title with tilt information does not get updated. If a different field is selected, the information is updated.

nguy:
Fixed in PR #170

PR #170, opened 2016-11-22
MAINT: Python 3 update.

nguy:
PEP8 corrections.

Python 3 updates that @gamaanderson put forth in earlier rescinded PR. In addition, some housekeeping with PEP8.


gamaanderson:
I think you can merge this without problem. If you don't get to it, I will merge in my local anyway, so there will be no conflicts.

nguy:
Merging in. I also added a fix for Issue #166 with this push. If anything needs updated, we can do that later.

Issue #164, opened 2016-10-01
Field calculation


gamaanderson:
I had an idea of how to use python magic to allow calculation and display of fields in a easy and intuitive way. something like:

```
reflectivity *= 2
```

 Just wanted to register it here so I don't forget.

nguy:
Interesting idea!

gamaanderson:
I have it here [https://github.com/gamaanderson/artview/tree/radar_terminal]

I made a small (sound less video) to show the idea
https://www.youtube.com/watch?v=jXVg3g8cnck

nguy:
That is pretty neat and seems like it could be relatively easy to implement?

gamaanderson:
it use pretty depp [python magic](https://docs.python.org/2/reference/datamodel.html#special-method-names), but it is easy. The only thing I don't like is that we still depend on the system terminal, emulating one is more complicated.

nguy:
I agree it's not perfect, but better than nothing for the time being. I guess the argument could be made either way: If the tools are in place, porting may be simpler OR if we get too many tools in place, then porting is going to be a pain. Hmmm.

gamaanderson:
this was closed by PR #175

PR #175, opened 2017-01-17
Radar terminal

gamaanderson:
PR #164 

give a look how this works, just as the video in the discussion. As before I still rely into the system terminal.


nguy:
I like the idea and I view this as an ongoing development topic. This PR should provide a template to begin. That said, I still find it a bit hard to use, probably because I haven't used it extensively. I would still like to see an easy GUI interface attached for simpler operations such as multiply or add a factor to a field.

nguy:
I'm happy merging this in as-is, it seems stable and extensible. And then we can continue to bulk this out as we go ahead. 

One thing that could be useful is brief example in the Help text mocking a terminal operation.

gamaanderson:
adding a constant (e.g. 10) is as easy as:

    reflectivity += 10

or

    new_reflectivity = reflectivity + 10
    add_field("new_reflectivity")

nguy:
Yes, I was able to easily do this following your tutorial. I was just saying that maybe we could add exactly what you posted into the Help as a brief and pertinent example.

gamaanderson:
just updated the help

nguy:
Alright, let's merge this to continue momentum.

PR #176, opened 2017-01-17
add correlation

gamaanderson:
show simple correlation graphic. It could use some expansion, including a recession line.

### DIFF
--- a/artview/components/correlation.py
+++ b/artview/components/correlation.py
@@ -1,0 +1,603 @@
+"""
+plot_radar.py
+
+Class instance used to make Display.
+"""
+# Load the needed packages
+import numpy as np
+import os
+import pyart
+
+from matplotlib.backends import pylab_setup
+backend = pylab_setup()[0]
+FigureCanvasQTAgg = backend.FigureCanvasQTAgg
+NavigationToolbar = backend.NavigationToolbar2QT
+#from matplotlib.backends.backend_qt4agg import FigureCanvasQTAgg
+#from matplotlib.backends.backend_qt4agg import NavigationToolbar2QT as \
+#    NavigationToolbar
+from matplotlib.figure import Figure
+from matplotlib.colors import Normalize as mlabNormalize
+from matplotlib.colorbar import ColorbarBase as mlabColorbarBase
+from matplotlib.pyplot import cm
+
+from ..core import (Variable, Component, common, VariableChoose, QtCore,
+                    QtGui, QtWidgets)
+from ..core.points import Points
+
+# Save image file type and DPI (resolution)
+IMAGE_EXT = 'png'
+DPI = 200
+# ========================================================================
+
+
+class Correlation(Component):
+    '''
+    Class to create a correlation plot, using a returned Radar structure
+    from the PyArt pyart.graph package.
+    '''
+
+    Vradar = None  #: see :ref:`shared_variable`
+    VfieldVertical = None  #: see :ref:`shared_variable`
+    VfieldHorizontal = None  #: see :ref:`shared_variable`
+    Vtilt = None  #: see :ref:`shared_variable`
+    Vgatefilter = None  #: see :ref:`shared_variable`
+    VplotAxes = None  #: see :ref:`shared_variable` (no internal use)
+
+    @classmethod
+    def guiStart(self, parent=None):
+        '''Graphical interface for starting this class'''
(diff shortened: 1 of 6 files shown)


gamaanderson:
I think I have forgot to add a mode for it, you may need to do the linking your self.

nguy:
With the merging of the radar terminal, there are now conflicts, can your rebase and push an update?

gamaanderson:
done

artview/components/plot_radar.py
```diff
@@ -44,6 +44,7 @@ class RadarDisplay(Component):
     Vgatefilter = None  #: see :ref:`shared_variable`
     VplotAxes = None  #: see :ref:`shared_variable` (no internal use)
     VpathInteriorFunc = None  #: see :ref:`shared_variable` (no internal use)
+    VpyartDisplay = None  #: see :ref:`shared_variable`
```
nguy: I was under the impression that you were trying to avoid creating a display as a shared variable?

gamaanderson: We first of all, it is already there, I have done this some time ago for the background, this line is only so that it appears in the docs (by the way it is been some time I have updates the docs)

second, this is not our display, but the pyart display class. I have never a problem with it, it was just redundant. Now it is needed because the display is what control the image coordinates. It also add the possibility on some future optimization where the radar values get changed but not the geometry.

gamaanderson: I continue to be agains classes accessing each other and I do like how the concept of shared variables forces the class to organize it self. I don't think this is something I will be changing in the future. That being said, I do see some space for increase interaction and coordination between the components. Modes are good, but they are still not perfect and have some performance issues.

artview/components/correlation.py
```diff
@@ -0,0 +1,630 @@
...
+        '''Create the Display Options Button menu.'''
+        parentdir = os.path.abspath(os.path.join(os.path.dirname(__file__),
+                                                 os.pardir))
+        config_icon = QtGui.QIcon(os.sep.join(
+            [parentdir, 'icons', "categories-applications-system-icon.png"]))
+        self.dispButton = QtWidgets.QPushButton(config_icon, "", self)
+        self.dispButton.setToolTip("Adjust display properties")
+        self.dispButton.setFocusPolicy(QtCore.Qt.NoFocus)
+        dispmenu = QtWidgets.QMenu(self)
+
```
nguy: It would be really useful to have controls on the data plot. Including scatter marker type, color, size, and maybe by default remove the edgecolor.

Also, making sure the the regression line bounds to the current axes limits so that the plot resize doesn't shrink the data of interest.

nguy:
I wonder if this feature would be better served as a data analysis general method where the user could choose a correlation or a historgram with stats printed out?

nguy:
Maybe combining with the select region and borrowing/sharing tools?

artview/components/correlation.py
gamaanderson: thank you for the sugestion, this should be not too dificult to implement. Could but be part of a second PR.

gamaanderson:
Yes, we could expand it to a data analysis, including use a `Points` container from the select region. However I would make so that it does not depend on a select region. That may need some switching, but all thinks we have done before.

gamaanderson:
Pyart is kind of poor in data analisis so we could gather those in a small module/file. That would include the 3 methods: `plot_correlation`, `plot_regression` and `_get_xy_values` (renamed). If you have any other plot function that can easily be put in this form we could also add.

artview/components/plot_radar.py
nguy: I wholeheartedly agree with everything you said. Let's keep an eye out for better ways in the future with working functionality now. Some performance I think we'll have to live with as a function of reasonably large data files (i.e. number of points).

nguy:
That was my thought about the `Points` container. The select region was just an example and I definitely don't want dependence there.

nguy:
I really like that idea of a small module or file. As with the other  PR, let's get this merged for the infrastructure beginning and go forward with further development. Could you start an Issue so we have a reminder for this?

gamaanderson:
the next issue to address is the layout. there is probably now I big conflict in the **modes.py** file that will need manual work. I will open an PR when I have it ready. Would you like a video showing it on my machine?

nguy:
That would be great!

PR #177, opened 2017-01-25
Layout

gamaanderson:
see video:
https://www.youtube.com/watch?v=BBeiI-6a5SI&feature=youtu.be


nguy:
Some general comments after playing with this PR a bit. It's a stream of things I want to get down and you can decide whether to address the issues or not. This is a big step toward Solo-like feel for many of us that are used that environment.

- [ ] Provide a list of possible layout options at the command line.
- [ ] Rename layout to panes or whatever when deploying. I think layout will be too general.
- [ ] I can't seem to control which view gets split when choosing a vertical or horizontal split. Or at least the highlighted pane is not always obvious.
- [ ] Better drag and drop? I don't think there is anything we can do here, but are limited by Qt.
- [ ] matplotlib figure does not scale nicely at times. Can we control this better? I'm not sure. The colorbar can be an issues especially when dealing with RHI plots. This could be a change that needs to happen in Py-ART.
- [ ] Just a thought. Make colorbar optional on the figure or show up in a separate pane?
- [ ] Bring back the right clicking performance that was there previously? This might help to add functionality and better user experience.
- [ ] Make an optional number of views, for example Solo had a drop down menu where you could choose 1x1, 2x2, 4x1 - maybe more, but I think this is enough.

nguy:
How is "close inactive views" determined? It would be nice to close any empty views. When I used this all of my windows went away except for the empty window.

gamaanderson:
Thank you for the feedback. I will address issue by issue.

1. The layout is built in run time, that is from outside of the classes. This mean we can use the current script section to offer different layouts 
2.  Do you mean renaming "views" to "panels". I like that.
3.  I'm aware of that, but I have not found a better solution. I could mark some panels as un-splittable and the actions in the layout menu get inactive then. 
4.  Yes, we are limited by Qt. But how would you like it improved?
5.  This is a separated issue and yes it is complicated to work with matplotlib rescaling, but we could/should try to improve that.
6.  That could be done, but is also separated issue.
7.  The problem with right click is that is complicated to know which widget will be receiving the signal. In such a dynamic environment I don't think we should trust this kind of user interation. But you could change my mind on this if you present a good concept for handling this.
8.  Good Idea, I would like to implement that, but there is a problem: consider that the user do a lot of slits and/or close panels and then he clicks 4x4. What happens then? Other question, should we set the top and left panels as permanent (not closeable)?
9.  "Close inactive views" means close all views that are not the current one. May be this is too drastic and we should change to "close empty views"

what do you think of moving the menu to the right, to avoid confusion with the file button under it?

artview/components/navigator.py
```diff
@@ -47,7 +465,7 @@ def __init__(self, Vradar=None, Vgrid=None, Vfilelist=None,
             Parent instance to associate to this class.
             If None, then Qt owns, otherwise associated w/ parent PyQt instance
         '''
-        super(FileNavigator, self).__init__(name=name, parent=parent)
+        super(FileNavigator_old, self).__init__(name=name, parent=parent)
         self.central_widget = QtWidgets.QWidget()
```
nguy: Just trying to get a hold here, what is the difference between the `FileNavigator` and `FileNavigator_old`

gamaanderson: just a copy I made before start changing everything. This and other things should be removed before merge

nguy:
I'll answer inline above to try to keep any confusion down.

nguy:
This is really an impressive amount of work you've put in @gamaanderson . Some really nice ideas in here.

gamaanderson:
I have not received any comments. If you made in form of revision, you have to submit the revision.

nguy:
I just added the comments to your previous post  above with and NG preceding my text.

gamaanderson:
ohn, i did not know you could do that! I will work on some of the suggestions and come back you.

gamaanderson:
Just pushed a new version, take a look. There is still a lot a debug features that I will remove before merge, ignore those for now.
Also added new comment to the post above.
What do you think, what must be done before merge? what can we address later?

gamaanderson:
one more comment: the window class is really complicated and scripts most work with it correctly. It is not foolproof at all. This really does not please me, but I decided to ignore this feeling in order to improve the GUI design. I'm also sorry it has almost zero documentation, but I would need some pages to really explain how the splitting tree works.

nguy:
A couple of initial comments:

1. For the command line parser, I think we could add more information. I could do this in a later PR if you want. Basically, I want to especially list the script options for startup. Looking at this, I think I see why the file open doesn't work well on a Windows machine!
2. When choosing 2x2 Display, the entire window is unable to be resized horizontally. This may have to do with our default sizing?
3. When choosing a mode and you have popped out the original window, it will appear there. Is this okay? Could be hard to locate.
4. I like the ARTview, Layout, Modes menu on the left. I think we are trained to look there first in software.
5. I really like how the Close button on the windows popped out closes just that window!

I may have more, let me keep testing.

gamaanderson:
Ok, Thank you. I will however only be able to adress that in the next week. I'm sure you know that, but just as a remark if you want to make code changes before merging, you can branch my branch to your account and add there commits. Then we make a reverse pull request.

gamaanderson:
1. ok, you can do it.
2. I image you mean re-size down. I try to set all kind of flag to allow re-size bellow minimum, but it didn't work. The only way to improve that is reduce the space each component is using.
3. Good catch. I solved it with some small ifs. But this indicates that the window and the modes are not working together. A solution would pull the window.py out of the /components, but I must think about it, I'm inclined to let it there for now.
4. ok. I putted a open and save text in the button, we have enough space in that line
5. that was not a conscious decision, but nice.

gamaanderson:
just pushed a new commit reverting the standard to classic (use now `-s layout` for the new), and did the clean up (as well as a proto logging).

If i didnt forget anything, what is possible I have, this should be able to merger.

nguy:
It looks good enough to merge for me. We can continue to work on it as we go.

PR #179, opened 2017-01-29
select region: update points when tilt changes

gamaanderson:
addresses old complain that points from select region get outdate by changing the sweep.

important discovery: function exported using shared variable can be update (i.e. call the `update` method) to reflect a change in the expected function results. This is really good.

Not sure if this is a strong or a weak update though, for now I set it to strong.

### DIFF
--- a/artview/components/plot_grid.py
+++ b/artview/components/plot_grid.py
@@ -499,6 +499,7 @@ class GridDisplay(Component):
             display = pyart.graph.GridMapDisplay(self.Vgrid.value)
             self.VpyartDisplay.change(display)
             self._update_infolabel()
+            self.VpathInteriorFunc.update(True)
 
     def NewField(self, variable, strong):
         '''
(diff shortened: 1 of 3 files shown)


nguy:
What are you using for a weak vs strong update definition?

nguy:
I'm just trying to follow along. Any shared variable should have the ability to be updated/changed the way the code is written? In this case the function that is called returns an updated shared variable, correct?

gamaanderson:
in general weak update indicate components not to perform too intensive calculations or that changes are only superficial, but in the end it is case specific, there is no general rule.

no, the function don't return a shared variable. The shared variable `.value` is a function. For instance `VpathInteriorFunc.value` is equal to `RadarDisplay.getPathInteriorValues`

nguy:
Aha, I see. In this case it is sort of grey whether it's strong or weak.

nguy:
Looks good. I did find another issue we should address at some point, I'll open a ticket.

PR #186, opened 2017-03-21
Issue181

gamaanderson:
deprecate display_select_region and add a select_region mode to display


nguy:
Could you rebase this. It has conflicts.

artview/components/plot_radar.py
```diff
@@ -438,16 +438,23 @@ def _add_toolsBoxUI(self):
...
         self.toolsButton.setMenu(toolmenu)
 
+    def add_mode(self, mode, label):
+        """ Add a tool entry with given label. Selecting that tool, execute
+        mode passing this component shared variables."""
+        def call_mode():
+            mode(self.get_sharedVariables())
+        action = self.toolmenu.addAction(label)
+        action.triggered.connect(call_mode)
+
```
nguy: I like the embedded functions. I was thinking about this the other day and it might be useful for this package.

nguy:
I just put up a short PR #191 , could you check briefly and hopefully we can merge that before you rebase. Thanks @gamaanderson !

artview/core/core.py
```diff
@@ -284,8 +284,15 @@ def closeEvent(self, QCloseEvent):
         self.disconnectAllVariables()
         super(Component, self).closeEvent(QCloseEvent)
 
+    def get_sharedVariables(self):
+        ''' get dictionaty of the shared variables '''
```
nguy: 'dictionaty' should be 'dictionary'

nguy:
The Menu window is not resizable, it looks like it is now based upon the number of tabs that are open. The window can be resized vertically, but not horizontally.

nguy:
I wonder if under Modes selection we should rename the following
'Query a selectable region of interest' to 'Region of Interest Polygon'
'Extract a selected region of points' to 'Query a Select Area'

I'm open to other suggestions.

nguy:
When I have the current `Extract a selected region of points` tool open I noticed an issue with the Plot Types.
 If I change tilts or files with the Histogram open, it replots without error. However, if I have the Table or Statistics open then the window splits vertically and the top never goes away. In fact if you repeat this it continues to split the window vertically adding an additional instance for each signal.

artview/components/plot_points.py
```diff
@@ -303,7 +303,18 @@ def NewPoints(self, variable, strong):
...
+                self.canvas.draw()
+            elif self.plot_type == 'statistics':
+                self.layout.removeWidget(self.statistics)
+                self.statistics.close()
+            elif self.plot_type == 'table':
+                self.layout.removeWidget(self.table)
+                self.table.close()
             return
 
         # Get field names
```
nguy: I think we may need to add an update method in the `core.common.CreateTable` class to redraw the values in the table. Also, it might be useful to also create a method to replace an individual cell. We could hook this up later to allow table editing.

Check out the following [SO link](http://stackoverflow.com/questions/24371274/qtextedit-how-to-dynamically-update-in-pyqt-without-crashing-gui) for the refreshing text.

Are the issues lies in `changePlotType` method, where it passes on statistics and table?

One more thought about the QTextEdit, could it be tied to the fact we set `setReadOnly` to True? Or does that just allow the user the ability to modify text?

gamaanderson:
thank you for the input. I will take a look in this tomorrow or Wednesday

gamaanderson:
the resizing is only being limited by the minimum size, this was already like this.

the easy result for the table is properly delete the previous one.

I would suggest:
'Query a selectable region of interest' to 'Select Polygons'
'Extract a selected region of points' to 'Query Selected Area'

nguy:
Hmm, I don't remember not being able to resize the windows. 

Whatever works for the table. 

You're wording is great!

gamaanderson:
commit pushed

nguy:
Wording is fixed and the transition between tilts/files now updates properly.

Could you roll the micro rev number for this PR, it's been a while and would be good at this point.

nguy:
I still don't get how the minimum size is controlling being able to expand a window? Especially since it only seems to be limiting horizontal expansion.

gamaanderson:
the size setting is complicated and I also believe it is kind of OS dependent. I'm being able to expand the widget without problems.

nguy:
Huh, really, well that is weird. MacOSX is going downhill fast.

gamaanderson:
I should get me acess to a MacOSX so that I can test it. I have however no idea where to get it. In the mean while, may be you could try te check older version to pin point where the problem started.

PR #188, opened 2017-03-21
Correlation

gamaanderson:
I just found this branch I had forgot about.

I have to take a look if it is all ok, but will open this PR to avoid forgetting.


gamaanderson:
ok, now this is good.

nguy:
Could you rebase this branch to resolve conflicts?

gamaanderson:
conflicts solved

artview/core/common.py
```diff
@@ -229,25 +229,37 @@ def handleClick(button):
 # Option methods #
 ##################
 
+
+def float_or_none(text):
+    if text=="None":
+        return None
+    else:
+        return float(text)
+
```
nguy: A possibly more robust test for the above could be something like
```
try:
    return float(text)
except:
    return None
```

This might handles cases of misspelling None or some other weird value or type

artview/components/correlation.py
```diff
@@ -570,10 +632,21 @@ def _update_plot(self):
         self.ax.set_xlabel(self.unitsHorizontal)
         self.ax.set_ylabel(self.unitsVertical)
 
+        print (self.parameters["xmin"], self.parameters["xmax"])
```
nguy: Remove space after print

nguy:
I think we should rename the Menu option to the correlation module `Regression Line` to `Linear Regression`. This is a bit of a nit, but it clearly conveys that a linear regression technique is used vs polynomial fitting for example.

nguy:
Could we also rename the main Modes entry from `Radar Moments Correlation` to `Field Correlation` or `Variable Correlation` or such. Since calculated fields are in the list as well as the base radar moments, this is more correct.

artview/core/common.py
gamaanderson: I actually want this function to raise an error in case of misuse, so that I can Handel this exception, for instance reset to previous value or standard value. I could add other spelling of `None`, but None is not to be a except value.

gamaanderson:
done

PR #191, opened 2017-03-24
Some minor fixes.

nguy:
One in particular could help with Windows compatibility, the use of the 'os' module where a unix specific designation was previously still used.


nguy:
One other thing. I was considering removing the bk_parser as it is legacy. Would you support that?

gamaanderson:
this look good and yes `bk_parser` can go.

PR #206, opened 2017-08-31
V1.3.8

gamaanderson:
correct bug with missing basemap


artview/components/correlation.py
```diff
@@ -642,12 +641,25 @@ def _update_plot(self):
...
-                str(self.horizontal_scale_menu_group.checkedAction().text()),
-                str(self.vertical_scale_menu_group.checkedAction().text()),
+                xscale,
+                yscale,
                 color=self.parameters["color"])
 
         self.canvas.draw()
 
+    def get_xyscale(self):
+        scale = [None, None]
```
nguy: Why bother with creating this list and then filling? Could just create the list.

artview/components/correlation.py
```diff
@@ -642,12 +641,25 @@ def _update_plot(self):
...
+        scale[0] = str(self.horizontal_scale_menu_group.checkedAction().text())
+        scale[1] = str(self.vertical_scale_menu_group.checkedAction().text())
+        
+        for i in range(2):
+            if scale[i] == "&linear":
+                scale[i] = "linear"
+            elif scale[i] == "l&og":
+                scale[i] = "log"
+        return scale
```
nguy: Will the `&` character always be in that position?
This [SO](https://stackoverflow.com/questions/3939361/remove-specific-characters-from-a-string-in-python) has a discussion of methods to remove an arbitrary string. Another discussion notes that even though the `.replace` method is techinically not the right way to do it, it is very fast.

```
scale[i].replace("&", "")
```

nguy:
Some issues I found when testing (in Python 3, Qt4).
  1. When I start the program, the Modes menu cannot be accessed until I make a change in the Menu Window (selected a SharedVariable). Further testing indicates that there is a time issue. If I do nothing and wait, I am eventually able to access the menu.
  2. The Menu and Displays became unlinked.?
  3. When choosing Edit Colormap, the FileNavigator Plugin also started. Also the Open File window is launched.
  4. Choosing the logarithmic colormap causes a fatal error the stops the application
```    
"Data has no positive values, and therefore can not be "
ValueError: Data has no positive values, and therefore can not be log-scaled.
Abort trap: 6
```
  5. The following error was output when choosing the `pyartDisplay` option in SharedVariables Window tab:
```
  File "/Users/guy/anaconda/envs/py35_qt5/lib/python3.5/site-packages/artview/components/plot_radar.py", line 616, in NewDisplay
    self._update_plot()
  File "/Users/guy/anaconda/envs/py35_qt5/lib/python3.5/site-packages/artview/components/plot_radar.py", line 981, in _update_plot
    plot_ppi = display.plot_ppi
AttributeError: 'NoneType' object has no attribute 'plot_ppi'
Abort trap: 6
```

